PR #4, opened 2019-12-14
Run miri in CI

Ralith:


CAD97:
You shouldn't need to manually install xargo; miri provides `cargo miri setup` specifically to do the first-time setup that would prompt for user input with just `cargo miri`.

PR #7, opened 2019-12-15
Use recursive macro to generate tuple bundles

cedric-h:


cedric-h:
@googlebot I signed it!

PR #10, opened 2019-12-15
Swap in fxhash for hashbrown, to begin reaching for #[no_std]

cedric-h:


Cargo.toml
```diff
@@ -10,20 +10,20 @@ readme = "README.md"
...
 
-[badges]
-maintenance = { status = "actively-developed" }
-
 [package.metadata.docs.rs]
 all-features = true
 
+[badges]
+maintenance = { status = "actively-developed" }
+
```
Ralith: Why was this moved?

PR #11, opened 2019-12-16
Add #[no_std] support

cedric-h:
wooooo

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ macros = ["hecs-macros", "lazy_static"]
 [dependencies]
 hecs-macros = { path = "macros", version = "0.1.0", optional = true }
 hashbrown = "0.6.3"
-lazy_static = { version = "1.4.0", optional = true, features = ["spin-no-std"] }
+lazy_static = { version = "1.4.0", optional = true, features = ["spin_no_std"] }
 
 [dev-dependencies]
 bencher = "0.1.5"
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -10,20 +10,22 @@ readme = "README.md"
 keywords = ["ecs", "entity"]
 categories = ["data-structures", "game-engines"]
 
-[badges]
-maintenance = { status = "actively-developed" }
-
```
Ralith: Why was this moved?

Cargo.toml
```diff
@@ -10,20 +10,22 @@ readme = "README.md"
...
 [features]
+default = ["std"]
+std = []
 # Enables derive(Bundle) and derive(Query)
 macros = ["hecs-macros", "once_cell"]
 
 [dependencies]
-fxhash = "0.2.1"
 hecs-macros = { path = "macros", version = "0.1.0", optional = true }
 once_cell = { version = "1.2", optional = true }
```
Ralith: This isn't no-std compatible.

Cargo.toml
```diff
@@ -10,20 +10,22 @@ readme = "README.md"
 keywords = ["ecs", "entity"]
 categories = ["data-structures", "game-engines"]
```
Ralith: Let's add `no-std` here to celebrate.

Cargo.toml
cedric-h: I was lazy and used `cargo add hashbrown`, and somewhere in the reading/writing process it decided to also move this.

PR #13, opened 2019-12-16
Re-expose query_scope

Ralith:


Ratysz:
An alternative:
```rust
    pub fn archetypes_len(&self) -> usize {
        self.archetypes.len()
    }

    pub fn archetype_in_scope<Q: Query>(&self, archetype: usize) -> bool {
        Q::Fetch::wants(&self.archetypes[archetype])
    }
```
(could use `u32` or most other unsigned ints instead, naturally)

This allows users to construct their own `query_scope()` implementations that minimize the amount of times `archetypes` needs to be iterated:
```rust
fn archetypes_of_one<Q: Query>(world: &World) -> impl Iterator<Item = usize> + '_ {
    (0..world.archetypes_len()).filter(move |&i| world.archetype_in_scope::<Q>(i))
}

fn archetypes_of_two<Q1: Query, Q2: Query>(world: &World) -> impl Iterator<Item = usize> + '_ {
    (0..world.archetypes_len())
        .filter(move |&i| world.archetype_in_scope::<Q1>(i) || world.archetype_in_scope::<Q2>(i))
}
```

One of my cases would use it in a macro:
```rust
    fn write_archetypes(world: &World, archetypes: &mut ArchetypeSet) {
        archetypes.extend((0..world.archetypes_len())
            .filter(move |&i| {
                $(world.archetype_in_scope::<$query>(i) ||)*
                false
            }))
    }
```

Ralith:
At that point I think it'd be clearer to just expose `&[Archetype]` with `impl Archetype { pub fn read_by<Q: Query>() -> bool; }`. Did you work out whether a distinct write case was useful?

Ratysz:
That would be clearer, yes; I was trying to avoid exposing any new types.

I don't think there needs to be a read/write distinction for archetypes. It's specified by the query on per-component level, so the caller of methods with `Query` parameter must know it already.

Ralith:
> I was trying to avoid exposing any new types.

Keeping the interface small is good, but making something more complex/obscure in pursuit of technically having fewer elements is another thing entirely. The ultimate goal here is obviousness.

> It's specified by the query on per-component level, so the caller of methods with Query parameter must know it already.

I think it's reasonable for hecs not to expect downstream tools to reimplement the query decoding logic. That's asking for drift, and increases the total amount of code in an end user's application.

Ratysz:
Was about to edit my comment; that might indeed reduce amount of code while improving granularity and performance of several disjointedness checks on my side, at very least.

Ralith:
Cool, I'll see about drafting the interface I sketched above, plus `fn written_by`.

Ralith:
Closing in favor of #51.

PR #30, opened 2020-01-12
Support reserving entity IDs concurrently

Ralith:


Ratysz:
This appears to fail on the third assertion.
```rust
#[test]
fn minimal_repro() {
    type MyQuery<'a> = (&'a usize, &'a f32);
    let mut world = hecs::World::new();
    world.spawn((0usize, 0f32));
    world.spawn((0usize, 0f32));
    let entities = world
        .query::<MyQuery>()
        .iter()
        .map(|(entity, _)| entity)
        .collect::<Vec<_>>();
    assert_eq!(entities.len(), 2);
    assert!(world.despawn(entities[0]).is_ok());
    let entities = world
        .query::<MyQuery>()
        .iter()
        .map(|(entity, _)| entity)
        .collect::<Vec<_>>();
    assert_eq!(entities.len(), 1);
}
```

Ralith:
Fixed and expanded tests to cover that case, thanks!

PR #32, opened 2020-01-13
Delete redundant helpers

Ralith:


Extrawurst:
Does this mean this code is not going to work anymore?
```
self.world
            .query::<(&MapPos, &Tick, &u32)>()
            .with::<TroopMarker>()
            .iter()....
```

Ralith:
My thinking is that you can just write `query::<With<TroopMarker, (...)>>`,
and having alternative ways to do that is unnecessary. Is that
substantially less convenient for you?

Extrawurst:
actually it makes sense, but tbh the docs of `query` are lacking those examples of how to actually simply combine the sorts of With and Without and it felt natural to chain those together. the docs of `with()` and `without()` are ripping a whole into how to get the same result without those helpers now.

PR #51, opened 2020-02-03
Refactor and expose query archetype access

Ralith:
CC @Ratysz. This needs review for semantic correctness which I'm too sleepy to give it at the moment.

### DIFF
--- a/src/archetype.rs
+++ b/src/archetype.rs
@@ -23,9 +23,13 @@ use core::ptr::{self, NonNull};
 use hashbrown::HashMap;
 
 use crate::borrow::AtomicBorrow;
-use crate::Component;
+use crate::query::Fetch;
+use crate::{Access, Component, Query};
 
 /// A collection of entities having the same component types
+///
+/// Accessing `Archetype`s is only required for complex dynamic scheduling. To manipulate entities,
+/// go through the `World`.
 pub struct Archetype {
     types: Vec<TypeInfo>,
     state: HashMap<TypeId, TypeState>,
@@ -287,6 +291,11 @@ impl Archetype {
             .as_ptr()
             .cast::<u8>();
         ptr::copy_nonoverlapping(component, ptr, size);
+    }
+
+    /// How, if at all, `Q` will access entities in this archetype
+    pub fn access<Q: Query>(&self) -> Option<Access> {
+        Q::Fetch::access(self)
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -69,16 +69,17 @@ mod entity_builder;
 mod query;
 mod world;
 
+pub use archetype::Archetype;
 pub use borrow::{EntityRef, Ref, RefMut};
 pub use bundle::{Bundle, DynamicBundle, MissingComponent};
 pub use entities::{Entity, NoSuchEntity};
 pub use entity_builder::{BuiltEntity, EntityBuilder};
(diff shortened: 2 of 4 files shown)


Ratysz:
```rust
#[test]
fn archetype_access() {
    let mut world = World::new();
    world.spawn(());
    world.spawn((0usize,));
    type NoComps<Q> = Without<usize, Q>;
    type OnlyUsize<Q> = With<usize, Q>;
    let mut archetypes = world.archetypes();

    let a = archetypes.next().unwrap();
    assert_eq!(a.access::<NoComps<()>>(), Some(Access::Iterate));
    assert_eq!(a.access::<OnlyUsize<()>>(), None);
    assert_eq!(a.access::<NoComps<Option<&usize>>>(), Some(Access::Iterate));

    let a = archetypes.next().unwrap();
    assert_eq!(a.access::<NoComps<()>>(), None);
    assert_eq!(a.access::<OnlyUsize<()>>(), Some(Access::Iterate));
    assert_eq!(a.access::<NoComps<Option<&usize>>>(), Some(Access::Iterate));
}
```
This fails on last assert, claiming it should be `None`. Should it?

Ralith:
> This fails on last assert, claiming it should be None. Should it?

I'm not sure what you mean by "claiming it should be None", but as written, I'd expect that assert to fail because `Without<usize, Q>` will never access an archetype that contains a `usize` for any `Q`.

Note that this test is fragile because the order of `archetypes` is not guaranteed.

Ratysz:
> Note that this test is fragile because the order of `archetypes` is not guaranteed.

I know; haven't focused on a more robust way of differentiating them yet. Should probably use the trivial asserts in a `match` instead.

Let me rewrite this, because after reading my prior comment I realised it says something completely different to what I actually meant. Commented-out assert fails here:

```rust
#[test]
fn archetype_access() {
    let mut world = World::new();
    world.spawn(());
    world.spawn((0usize,));
    type NoComps<Q> = Without<usize, Q>;
    type OnlyUsize<Q> = With<usize, Q>;
    let mut archetypes = world.archetypes();

    let a = archetypes.next().unwrap();
    assert_eq!(a.access::<NoComps<()>>(), Some(Access::Iterate))
    assert_eq!(a.access::<OnlyUsize<()>>(), None);
    //assert_eq!(a.access::<NoComps<Option<&usize>>>(), None);
    assert_eq!(a.access::<OnlyUsize<Option<&usize>>>(), None);

    let a = archetypes.next().unwrap();
    assert_eq!(a.access::<NoComps<()>>(), None);
    assert_eq!(a.access::<OnlyUsize<()>>(), Some(Access::Iterate));
    assert_eq!(a.access::<NoComps<Option<&usize>>>(), None);
    assert_eq!(a.access::<OnlyUsize<Option<&usize>>>(), Some(Access::Read));
}
```

Ralith:
A robust test could construct a world that necessarily has exactly one archetype. You could also separately test the number of archetypes you get from a world with multiple (note that `archetypes` returns an `ExactSizeIterator`, which has a `len` method).

> assert_eq!(a.access::<NoComps<Option<&usize>>>(), None);

This assert should fail. `Without<usize, Option<&usize>>` will succeed vs. the empty archetype because the empty archetype does not contain a `usize`, therefore the outer clause is satisfied, and `Option<Q>` will `Iterate` archetypes that do not satisfy `Q`.

Ralith:
On second thought, it's difficult to do interesting tests on worlds with only one archetype because every world has the empty archetype.

PR #55, opened 2020-03-08
Use cargo miri setup rather than installing xargo manually

Ralith:
CC @CAD97


Ralith:
Not actually sure if this makes a material difference, but might as well.

CAD97:
Mainly it's just a "say what you want not how to get it" kind of thing.

Plus, if miri ever starts needing some other setup (or stops needing xargo), asking miri to do the setup will gracefully update whereas just installing xargo will just silently stop working.

Ralith:
Makes sense!

Issue #50, opened 2020-02-01
get multiple components


Extrawurst:
is there a way to `world.get` multiple components of an entity? to not pay the entity.get cost multiple times?

Ralith:
It's unlikely to matter for performance in practice, but [`World::entity`](https://docs.rs/hecs/0.2.7/hecs/struct.World.html#method.entity) provides that.

Extrawurst:
I see! Thanks for the hint at `World::entity`!

PR #58, opened 2020-03-14
Support querying a single entity

Ralith:
Still considering whether this is worth having.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -67,6 +67,7 @@ mod bundle;
 mod entities;
 mod entity_builder;
 mod query;
+mod query_one;
 mod world;
 
 pub use archetype::Archetype;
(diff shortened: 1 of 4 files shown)


1tgr:
Would be good to have this, just now I wrote a sequence of calls to `get(entity).unwrap()` for individual components.

extrawurst:
hm thought there is something for that (see https://github.com/Ralith/hecs/issues/50#issuecomment-581072732):

> It's unlikely to matter for performance in practice, but [`World::entity`](https://docs.rs/hecs/0.2.7/hecs/struct.World.html#method.entity) provides that.

1tgr:
Agree that `World::entity` allows me to access a single entity, but I still have to issue multiple calls to `get` or `get_mut` afterward. With this `World::query_one` it's a one-liner regardless of the number of components involved.

I'm not so concerned about performance in this case as convenience... the `query_` methods nicely handle the component checks for me, and they extend to `with`, `without` and `Option<T>` for more complex cases.

extrawurst:
Agreed, I think it would be a nice addition to have `query_one`

Ralith:
Thanks for the feedback! I'll move this forward. My only reservation was reluctance to increase the API surface area without anyone having actually asked for it.

Ralith:
Revised the API a bit for more consistency with `query`; in particular `with`/`without` work, the error cases can be distinguished, and there's one less layer of indirection. Any final feedback?

1tgr:
LGTM!

Ralith:
Released in 0.2.11.

PR #65, opened 2020-06-24
Added my own example that explores the use of hecs

Winsalot:
I had found the original example to be quite lacking in explaining how to use this ecs library. After taking a dive into docs I have written a simple simulation that has a lot of enitites interacting with each other.

I believe that this example would help some people unfamiliar with ECS to dive into using hecs faster.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,6 @@ lazy_static = { version = "1.4.0", optional = true, features = ["spin_no_std"] }
 bencher = "0.1.5"
 rand = "0.7.3"
 
-
 [[bench]]
 name = "bench"
 harness = false
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,8 @@ lazy_static = { version = "1.4.0", optional = true, features = ["spin_no_std"] }
 
 [dev-dependencies]
 bencher = "0.1.5"
+rand = "0.7.3"
+
 
 [[bench]]
 name = "bench"
--- a/examples/ffa_simulation.rs
+++ b/examples/ffa_simulation.rs
@@ -6,210 +6,219 @@ use std::io;
  Simple simulation
  Spawn multiple entities. They have health, damage, position and other components.
  On every tick every entity/unit:
-	1. Moves in random direction.
- 	2. Finds closest entity to itself.
- 	3. Fires at it and applies damage.
- 	4. Gets damaged by other entities firing at them.
- 	5. If health <= 0, the unit dies.
+    1. Moves in random direction.
+     2. Finds closest entity to itself.
+     3. Fires at it and applies damage.
+     4. Gets damaged by other entities firing at them.
+     5. If health <= 0, the unit dies.
 State of the simulation is displayed in the sconsole through println! functions.
 */
 
 #[derive(Debug)]
 struct Position {
-	x: i32,
-	y: i32,
+    x: i32,
+    y: i32,
 }
 
 #[derive(Debug)]
 struct Health {
-	hp: i32
+    hp: i32,
 }
 
 #[derive(Debug)]
 struct Speed {
-	speed: i32
+    speed: i32,
 }
 
 #[derive(Debug)]
 struct Damage {
-	dmg: i32
+    dmg: i32,
 }
 
 #[derive(Debug)]
 struct KillCount {
-	count: i32
+    count: i32,
 }
 
-
-fn manhattan_dist(x0: i32, x1: i32, y0: i32, y1: i32) -> i32{
-	let dx = (x0 - x1).max(x1 - x0);
-	let dy = (y0 - y1).max(y1 - y0);
-	return dx + dy
+fn manhattan_dist(x0: i32, x1: i32, y0: i32, y1: i32) -> i32 {
+    let dx = (x0 - x1).max(x1 - x0);
+    let dy = (y0 - y1).max(y1 - y0);
+    return dx + dy;
 }
 
 // Spawns entities one by one. It's better to use spawn_batch instead, which is used in function below.
-fn _spawn_enitites(world: &mut World, n: usize){
+fn _spawn_enitites(world: &mut World, n: usize) {
+    let mut rng = thread_rng();
 
-	let mut rng = thread_rng();
+    for _ in 0..n {
+        let pos = Position {
+            x: rng.gen_range(-10, 10),
+            y: rng.gen_range(-10, 10),
+        };
+        let s = Speed {
+            speed: rng.gen_range(1, 5),
+        };
+        let hp = Health {
+            hp: rng.gen_range(30, 50),
+        };
+        let dmg = Damage {
+            dmg: rng.gen_range(1, 10),
+        };
+        let kc = KillCount { count: 0 };
 
-	for _ in 0..n {
-		let pos = Position{
-			x: rng.gen_range(-10, 10),
-			y: rng.gen_range(-10, 10),
-		};
-		let s =  Speed{speed: rng.gen_range(1, 5)};
-		let hp = Health{hp: rng.gen_range(30, 50)};
-		let dmg = Damage{dmg: rng.gen_range(1, 10)};
-		let kc = KillCount{count: 0};
-
-		world.spawn((pos, s, hp, dmg, kc));
-	}
+        world.spawn((pos, s, hp, dmg, kc));
+    }
 }
 
-fn batch_spawn_entities(world: &mut World, n: usize){
+fn batch_spawn_entities(world: &mut World, n: usize) {
+    let mut rng = thread_rng();
 
-	let mut rng = thread_rng();
+    let to_spawn = (0..n).map(|_| {
+        let pos = Position {
+            x: rng.gen_range(-10, 10),
+            y: rng.gen_range(-10, 10),
+        };
+        let s = Speed {
+            speed: rng.gen_range(1, 5),
+        };
+        let hp = Health {
+            hp: rng.gen_range(30, 50),
+        };
+        let dmg = Damage {
+            dmg: rng.gen_range(1, 10),
+        };
+        let kc = KillCount { count: 0 };
 
-	let to_spawn = (0..n).map(|_|{
+        (pos, s, hp, dmg, kc)
+    });
 
-		let pos = Position{
-			x: rng.gen_range(-10, 10),
-			y: rng.gen_range(-10, 10),
-		};
-		let s =  Speed{speed: rng.gen_range(1, 5)};
-		let hp = Health{hp: rng.gen_range(30, 50)};
-		let dmg = Damage{dmg: rng.gen_range(1, 10)};
-		let kc = KillCount{count: 0};
-
-		(pos, s, hp, dmg, kc)
-	});
-
-	world.spawn_batch(to_spawn);
+    world.spawn_batch(to_spawn);
 }
 
-fn system_integrate_motion(world: &mut World){
+fn system_integrate_motion(world: &mut World) {
+    let mut rng = thread_rng();
 
-	let mut rng = thread_rng();
-
-	for (id, (pos, s)) in &mut  world.query::<(&mut Position, &Speed)>() {
-		let change = (
-			rng.gen_range(-s.speed, s.speed),
-			rng.gen_range(-s.speed, s.speed)
-			);
-		pos.x += change.0;
-		pos.y += change.1;
-		println!("Unit {:?} moved to {:?}",id, pos);
-	}
-
+    for (id, (pos, s)) in &mut world.query::<(&mut Position, &Speed)>() {
+        let change = (
+            rng.gen_range(-s.speed, s.speed),
+            rng.gen_range(-s.speed, s.speed),
+        );
+        pos.x += change.0;
+        pos.y += change.1;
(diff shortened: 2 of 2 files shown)


examples/ffa_simulation.rs
```diff
@@ -0,0 +1,208 @@
...
+		let dmg = Damage{dmg: rng.gen_range(1, 10)};
+		let kc = KillCount{count: 0};
+
+		(pos, s, hp, dmg, kc)
+	});
+
+	world.spawn_batch(to_spawn);
+}
+
+fn move_system(world: &mut World){
```
Ralith, line 89: Maybe name this `integrate_motion` or something? Having the `_system` suffix on only one of the "systems" is a bit odd.

Winsalot:
Hi. I have renamed all "system" functions to be `system_[function name]`, and added some other minor changes.

Ralith:
Thanks! Please address the formatting issues by running `cargo fmt` or `rustfmt`.

Winsalot:
Well, that's a new feature for me :smiley: . Done :+1:

Cargo.toml
```diff
@@ -29,6 +29,8 @@ lazy_static = { version = "1.4.0", optional = true, features = ["spin_no_std"] }
 
 [dev-dependencies]
 bencher = "0.1.5"
+rand = "0.7.3"
+
```
Ralith, line 33: Stray newline

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,224 @@
...
+// In this system entities find the closest entity and fire at them
+fn system_fire_at_closest(world: &mut World) {
+    for (id0, (pos0, dmg0, kc0)) in
+        &mut world.query::<With<Health, (&Position, &Damage, &mut KillCount)>>()
+    {
+        let mut min_dist: Option<i32> = None;
+        let mut closest_id: Option<Entity> = None;
+
+        // find closest:
+        for (id1, pos1) in &mut world.query::<With<Health, &Position>>() {
```
Ralith, line 121: It's worth leaving a comment here that explains that nested queries are O(n^2) and you usually want to avoid that by using some sort of spatial index like a quadtree or more general BVH, which we don't bother with here since it's out of scope for the example.

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,224 @@
...
+    count: i32,
+}
+
+fn manhattan_dist(x0: i32, x1: i32, y0: i32, y1: i32) -> i32 {
+    let dx = (x0 - x1).max(x1 - x0);
+    let dy = (y0 - y1).max(y1 - y0);
+    return dx + dy;
+}
+
+// Spawns entities one by one. It's better to use spawn_batch instead, which is used in function below.
```
Ralith, line 49: "better" is vague; consider being more explicit that `spawn_batch` has higher performance.

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,224 @@
...
+            return;
+        }
+
+        // Deal damage:
+        /*
+                //Get target unit hp like this:
+                let mut hp1 = world.query_one::<&mut Health>(closest_id.unwrap()).unwrap();
+                let hp1 = hp1.get().unwrap();
+        */
+        //Or like this:
```
Ralith, line 149:
```suggestion
        // Or like this:
```

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        if !closest_id.is_some() {
+            println!("{:?} is the last survivor!", id0);
+            return;
+        }
+
+        // Deal damage:
+        /*
+                //Get target unit hp like this:
```
Ralith, line 145:
```suggestion
                // Get target unit hp like this:
```

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,224 @@
+use hecs::*;
+use rand::{thread_rng, Rng};
+use std::io;
+
+/*
+ Simple simulation
+ Spawn multiple entities. They have health, damage, position and other components.
+ On every tick every entity/unit:
+    1. Moves in random direction.
```
Ratysz, line 9:
```suggestion
     1. Moves in random direction.
```

Ralith:
@Ratysz, you signed the CLA, right?

Ratysz:
Just did, yes.

Ratysz:
@googlebot I fixed it.

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,224 @@
...
+
+#[derive(Debug)]
+struct Damage {
+    dmg: i32,
+}
+
+#[derive(Debug)]
+struct KillCount {
+    count: i32,
+}
```
ducaale, line 41: Have you considered using tuple structs?

```rust
struct Health(i32);
struct Damage(i32);
struct KillCount(i32);
```

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    for i in 0..to_remove.len() {
+        world.despawn(to_remove[i]).unwrap();
+    }
+}
+
+fn print_world_state(world: &mut World) {
+    for (id, (hp, pos, dmg, kc)) in &mut world.query::<(&Health, &Position, &Damage, &KillCount)>()
+    {
+        println!("Entity stats:");
```
ducaale, line 192: moving this print statement outside of the loop would make the output much cleaner IMHO

```
Entity stats:
ID: 4v0, Health { hp: 39 }, Damage { dmg: 9 }, Position { x: -15, y: 0 }, KillCount { count: 0 }
ID: 1v0, Health { hp: 41 }, Damage { dmg: 8 }, Position { x: 2, y: 3 }, KillCount { count: 0 }
ID: 2v0, Health { hp: 39 }, Damage { dmg: 2 }, Position { x: -14, y: -5 }, KillCount { count: 0 }
ID: 3v0, Health { hp: 30 }, Damage { dmg: 7 }, Position { x: 0, y: 7 }, KillCount { count: 1 }
```

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+fn main() {
+    let mut world = World::new();
+
+    // _spawn_enitites(&mut world, 5);
+    batch_spawn_entities(&mut world, 5);
+
+    'running: loop {
+        println!("'Enter' to continue simulation, '?' for enity list, 'q' to quit");
```
ducaale, line 204: adding a new at the start would make a visually pleasing output

```
Unit 2v0 was damaged by 0v0 for 6 HP
Unit 0v0 was damaged by 1v0 for 8 HP
Unit 0v0 was damaged by 2v0 for 2 HP
Unit 0v0 was damaged by 3v0 for 7 HP
Unit 0v0 was killed by unit 3v0!

'Enter' to continue simulation, '?' for enity list, 'q' to quit
```

examples/ffa_simulation.rs
ducaale, line 192: adding a new at the start would make a visually pleasing output

```
'Enter' to continue simulation, '?' for enity list, 'q' to quit
?

Entity stats:
ID: 4v0, Health { hp: 39 }, Damage { dmg: 9 }, Position { x: -15, y: 0 }, KillCount { count: 0 }
ID: 1v0, Health { hp: 41 }, Damage { dmg: 8 }, Position { x: 2, y: 3 }, KillCount { count: 0 }
ID: 2v0, Health { hp: 39 }, Damage { dmg: 2 }, Position { x: -14, y: -5 }, KillCount { count: 0 }
```

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,224 @@
...
+fn system_remove_dead(world: &mut World) {
+    // Here we query entities with 0 or less hp and despawn them
+    let mut to_remove: Vec<Entity> = Vec::new();
+    for (id, hp) in &mut world.query::<&Health>() {
+        if hp.hp <= 0 {
+            to_remove.push(id);
+        }
+    }
+
+    for i in 0..to_remove.len() {
```
ducaale, line 184: would a normal iteration work here?

```rust
for id in to_remove {
    world.despawn(id).unwrap();
}
```

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,224 @@
...
+        pos.x += change.0;
+        pos.y += change.1;
+        println!("Unit {:?} moved to {:?}", id, pos);
+    }
+}
+
+// In this system entities find the closest entity and fire at them
+fn system_fire_at_closest(world: &mut World) {
+    for (id0, (pos0, dmg0, kc0)) in
+        &mut world.query::<With<Health, (&Position, &Damage, &mut KillCount)>>()
```
ducaale, line 115: Ok, this is a question from my side rather than a review comment. How does `With` works generally?

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,224 @@
...
+            hp: rng.gen_range(30, 50),
+        };
+        let dmg = Damage {
+            dmg: rng.gen_range(1, 10),
+        };
+        let kc = KillCount { count: 0 };
+
+        (pos, s, hp, dmg, kc)
+    });
+
```
ducaale, line 94: I think you can write the inefficient version here as a comment to avoid having an unused function laying around.

```rust
// some explanation to why this should be avoided
// for entity in to_spawn {
//     world.spawn(entity);
// }
```

examples/ffa_simulation.rs
Ralith: `With` selects only queries that have a certain component, but does not actually borrow that component.

ducaale: Ok, got it. Thanks!

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+#[derive(Debug)]
+struct KillCount {
+    count: i32,
+}
+
+fn manhattan_dist(x0: i32, x1: i32, y0: i32, y1: i32) -> i32 {
+    let dx = (x0 - x1).max(x1 - x0);
+    let dy = (y0 - y1).max(y1 - y0);
```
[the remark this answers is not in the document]
Winsalot: Added changes

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,224 @@
...
+fn system_fire_at_closest(world: &mut World) {
+    for (id0, (pos0, dmg0, kc0)) in
+        &mut world.query::<With<Health, (&Position, &Damage, &mut KillCount)>>()
+    {
+        let mut min_dist: Option<i32> = None;
+        let mut closest_id: Option<Entity> = None;
+
+        // find closest:
+        for (id1, pos1) in &mut world.query::<With<Health, &Position>>() {
+            if id0 != id1 {
```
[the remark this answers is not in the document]
Winsalot: Good idea

examples/ffa_simulation.rs
Winsalot: In my experience with normal iterations you can run into problems with burrow checker sometimes, therefore I try to use indices and not think about it.

Ralith: Normal iteration is preferred when it's possible. In this case we're iterating through a collection that's `'static` which is totally fine.

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,230 @@
...
+fn system_fire_at_closest(world: &mut World) {
+    for (id0, (pos0, dmg0, kc0)) in
+        &mut world.query::<With<Health, (&Position, &Damage, &mut KillCount)>>()
+    {
+        let mut min_dist: Option<i32> = None;
+        let mut closest_id: Option<Entity> = None;
+
+        //Find closest:
+        //Comment about implementation: Nested queries are O(n^2) and you usually want to avoid that by using some sort of spatial index like a quadtree or more general BVH, which we don't bother with here since it's out of scope for the example.
+        'child_loop: for (id1, pos1) in &mut world.query::<With<Health, &Position>>() {
```
ducaale, line 122: Is this label necessary?

examples/ffa_simulation.rs
```diff
@@ -0,0 +1,230 @@
...
+                }
+                Some(mut _dist0) => {
+                    _dist0 = _dist0.min(dist);
+                }
+            }
+
+            if Some(dist) == min_dist {
+                closest_id = Some(id1);
+            }
+        }
```
ducaale, line 141: I think there is a simpler way which avoids mutation and unwrapping, Not sure if it is more clearer though

```rust
for (id0, (pos0, dmg0, kc0)) in &mut world.query::<With<Health, (&Position, &Damage, &mut KillCount)>>() {
    // find closest:
    let mut entities = world.query::<With<Health, &Position>>();
    let closest = entities.iter()
        .filter(|(id1, _)| *id1 != id0)
        .min_by_key(|(_, pos1)| manhattan_dist(pos0.x, pos1.x, pos0.y, pos1.y));

     match closest {
         Some(closest) => {
             let (id1, _) = closest;
             // health code
         }
         None => {
             println!("{:?} is the last survivor!", id0);
         }
}
```

Ralith:
Added a few tweaks manually and merged in 6473bd04b0038e8b7639721529bbeb144e4bc1ad. Thanks!

Issue #71, opened 2020-08-15
Upstreaming Bevy ECS Changes


cart:
My recently announced [Bevy ECS](https://bevyengine.org/news/introducing-bevy/) project uses a forked version of hecs, which adds the following features on top:

(copied directly from the blog post)

* **Function Systems**: Hecs actually has no concept of a "system" at all. You just run queries directly on the World. Bevy adds the ability to define portable, schedulable systems using normal Rust functions.
* **Resources**: Hecs has no concept of unique/global data. When building games, this is often needed. Bevy adds a `Resource` collection and resource queries
* **Parallel Scheduler**: Hecs is single threaded, but it was designed to allow parallel schedulers to be built on top. Bevy ECS adds a custom dependency-aware scheduler that builds on top of the "Function Systems" mentioned above.
* **Optimization**: Hecs is already plenty fast, but by modifying some of its internal data access patterns, we were able to improve performance significantly. This moved it from "fast enough" to "the fastest" (see the benchmark above to compare Bevy ECS to vanilla Hecs).
* **Query Wrappers**: The `Query` Bevy ECS exports is actually a wrapper around Hecs Queries. It provides safe, scoped access to the `World` in a multi-threaded context and improves the ergonomics of iteration.
* **Change Detection**: Automatically (and efficiently) tracks component add/remove/update operations and exposes them in the Query interface.
* **Stable Entity IDs**: Almost every ECS (including Hecs) uses unstable entity ids that cannot be used for serialization (scenes / save files) or networking. In Bevy ECS, entity ids are globally unique and stable. You can use them in any context!  

Most of the performance improvements came from removing Entity from the iterator and instead returning it via queries. I suspect it allowed rust to inline something (or otherwise optimize it) in a way that it couldn't before. I also tweaked manual inlining in a few places.

I'd also like to note that Stable Entity IDs do come at a performance cost, and the current version uses random u32's, which bevy users have [proven to have a high collision risk](https://github.com/bevyengine/bevy/issues/117). That being said the collision risk is a solvable problem and the performance cost is "worth it" to me. (although entity ids are still a hot discussion, as you can probably see in that thread). You can see the difference illustrated in the performance graphs in the blog post above.

I'd love to upstream whatever features you want from Bevy ECS, so its mainly a question of "what do you want" / "what fits the hecs scope".

[...]

PR #72, opened 2020-08-15
Pull changes from bevy_hecs

multun:
The history is still very messy and will get plenty of rework: commits will get squashed, splitted, until the result looks good

EDIT: Blocked by #71, we need to decide what needs to get in before splitting this


multun:
@googlebot I signed it!

Ralith:
Closing as no longer applicable, especially as bevy has diverged. Thanks for your interest, and feel free to reopen if there are specific changes you think would still make sense for us!

PR #74, opened 2020-09-04
Allow clippy::forget_copy in generated code

toasteater:
The code generated by `derive(Bundle)` triggers a clippy correctness lint in user crate when one of the components in the bundle is `Copy`. This silences the lint.


toasteater:
I'm not feeling like exposing my personal email and phone number for a one-liner PR, so I'm closing this. Maybe commit this on your own @Ralith if you think it makes sense!

Ralith:
You don't need to expose anything that isn't in the commit metadata regardless, FWIW.

PR #82, opened 2020-10-09
EntityBuilder::add_bundle, swap to `TypeInfo` w/ `DynamicBundle::put` and `Bundle::get`

sdleffler:


sdleffler:
@googlebot I signed it!

src/entity_builder.rs
```diff
@@ -53,38 +53,98 @@ impl EntityBuilder {
...
+                unsafe {
+                    let storage_ptr = self
+                        .storage
+                        .as_mut_ptr()
+                        .cast::<u8>()
+                        .add(offset)
+                        .cast::<T>();
+
+                    // alloc a properly aligned tmp buffer and copy in the old value
+                    // so we can drop it safely
```
Ralith, line 75: Outdated comment.

src/entity_builder.rs
```diff
@@ -53,38 +53,98 @@ impl EntityBuilder {
...
+                        let index = *occupied.get();
+                        let (ty, offset) = self.info[index];
+
+                        let storage_ptr = self.storage.as_mut_ptr().cast::<u8>().add(offset);
+                        // alloc a properly aligned tmp buffer and copy in the old value
+                        // so we can drop it safely
+                        let tmp = alloc(ty.layout());
+                        ptr::copy_nonoverlapping(storage_ptr, tmp, ty.layout().size());
+                        ty.drop(tmp);
+                        dealloc(tmp, ty.layout());
```
Ralith, line 117: The fiddly dance here makes me think we should probably just store things properly aligned in the first place. Then I think there'd be no reason for `add` not to dispatch directly to `add_bundle`, which saves a lot of duplication. If you prefer we can leave that for future work, but I think it'd be a substantial benefit.

src/entity_builder.rs
```diff
@@ -53,38 +53,98 @@ impl EntityBuilder {
             storage: Box::new([]),
             info: Vec::new(),
             ids: Vec::new(),
-            id_set: HashSet::new(),
+            indices: HashMap::new(),
         }
     }
 
     /// Add `component` to the entity
```
Ralith, line 60: Let's document the behavior when a component already exists for both versions.

tests/tests.rs
```diff
@@ -130,6 +130,20 @@ fn build_entity() {
...
+    let mut world = World::new();
+    let mut entity = EntityBuilder::new();
+    entity.add_bundle(("abc", 123));
+    let e = world.spawn(entity.build());
+    entity.add_bundle(("def", [0u8; 1024], 456));
+    let f = world.spawn(entity.build());
+    assert_eq!(*world.get::<&str>(e).unwrap(), "abc");
+    assert_eq!(*world.get::<i32>(e).unwrap(), 123);
+    assert_eq!(*world.get::<&str>(f).unwrap(), "def");
+    assert_eq!(*world.get::<i32>(f).unwrap(), 456);
```
Ralith, line 144: Let's cover the case of replacing an existing component as well.

Ralith (changes requested):
Thanks! I'd prefer to not have commented out code committed, since it'll inevitably bitrot and may lose relevance.

src/entity_builder.rs
sdleffler: Added docs for this in the commit I just pushed; please take a look and see if they're adequate.

tests/tests.rs
sdleffler: Added! Please take a look.

src/entity_builder.rs
sdleffler: It shouldn't be difficult to do that and I agree that it would be good for `add` to dispatch directly to `add_bundle`. I'll look at making that improvement but I think it's substantial enough for me that I'd like to get this PR in first before I work on it.

src/bundle.rs
```diff
@@ -25,6 +25,9 @@ pub trait DynamicBundle {
     /// Invoke a callback on the fields' type IDs, sorted by descending alignment then id
     #[doc(hidden)]
     fn with_ids<T>(&self, f: impl FnOnce(&[TypeId]) -> T) -> T;
+    // /// Invoke a callback on the fields' type IDs, sorted by descending alignment then id
+    // #[doc(hidden)]
+    // fn with_infos<T>(&self, f: impl FnOnce(&[TypeInfo]) -> T) -> T;
```
Ralith, line 30: I'd prefer for this and the related commented dead code to be removed.

Ralith (changes requested):

src/bundle.rs
sdleffler: Oops! Didn't realize I'd left that in.

Ralith (approved):

PR #84, opened 2020-10-09
Archetype::component_types

sdleffler:
This PR adds a method with similar functionality to `Entity::component_types`, but on `Archetype`. This is useful for automating generation of scripting interfaces and schedules from archetypes.

### DIFF
--- a/src/archetype.rs
+++ b/src/archetype.rs
@@ -137,9 +137,17 @@ impl Archetype {
         &self.types
     }
 
-    /// Enumerate the types of the components of entities stored in this archetype
+    /// Enumerate the types of the components of entities stored in this archetype.
     ///
     /// Convenient for dispatching logic which needs to be performed on sets of type ids.
+    /// For example, suppose you're building a scripting system, and you want to integrate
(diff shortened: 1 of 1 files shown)


src/archetype.rs
```diff
@@ -137,6 +137,13 @@ impl Archetype {
         &self.types
     }
 
+    /// Enumerate the types of the components of entities stored in this archetype
+    ///
+    /// Convenient for dispatching logic which needs to be performed on sets of type ids.
```
Ralith, line 142: Could we add some more specific discussion of when this is useful? A brief summary of a specific application or similar can be a big help for both helping other users discover good patterns, and for ensuring future refactoring preserves the desired capabilities.

Ralith (changes requested):

src/archetype.rs
sdleffler: Done! I tried to describe the use case which motivated me to ask about this functionality, without getting into the specifics.

Ralith (approved):

PR #85, opened 2020-10-09
Add `World::resolve_unknown_gen`

sdleffler:
This adds an unsafe method `World::resolve_unknown_gen` which allows the user to map a `u32` identifier extracted via `Entity::id` back to a potentially live `Entity`. As discussed previously it's unsafe because it may return a reference to a dead `Entity`, which will appear as though it has no components and may later spontaneously acquire components by becoming a live `Entity`.

### DIFF
--- a/src/entities.rs
+++ b/src/entities.rs
@@ -236,6 +236,22 @@ impl Entities {
         Ok(meta.location)
     }
 
+    /// Returns `None` if the given id would represent an index outside of `meta`.
+    pub unsafe fn resolve_unknown_gen(&self, id: u32) -> Option<Entity> {
+        let meta_len = self.meta.len();
+        if meta_len + self.pending.load(Ordering::Relaxed) as usize <= id as usize {
+            None
+        } else if meta_len <= id as usize {
+            Some(Entity { generation: 0, id })
+        } else {
+            let meta = &self.meta[id as usize];
+            Some(Entity {
+                generation: meta.generation,
+                id,
+            })
+        }
+    }
+
     /// Allocate space for and enumerate pending entities
     #[allow(clippy::reversed_empty_ranges)]
     pub fn flush(&mut self) -> impl Iterator<Item = u32> {
--- a/src/entities.rs
+++ b/src/entities.rs
@@ -237,6 +237,9 @@ impl Entities {
     }
 
     /// Returns `None` if the given id would represent an index outside of `meta`.
+    ///
+    /// # Safety
+    /// Must only be called for currently allocated `id`s.
     pub unsafe fn resolve_unknown_gen(&self, id: u32) -> Option<Entity> {
         let meta_len = self.meta.len();
         if meta_len + self.pending.load(Ordering::Relaxed) as usize <= id as usize {
--- a/src/entities.rs
+++ b/src/entities.rs
@@ -240,18 +240,18 @@ impl Entities {
     ///
     /// # Safety
     /// Must only be called for currently allocated `id`s.
-    pub unsafe fn resolve_unknown_gen(&self, id: u32) -> Option<Entity> {
+    pub unsafe fn resolve_unknown_gen(&self, id: u32) -> Entity {
         let meta_len = self.meta.len();
         if meta_len + self.pending.load(Ordering::Relaxed) as usize <= id as usize {
-            None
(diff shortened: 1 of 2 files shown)


src/entities.rs
```diff
@@ -236,6 +236,22 @@ impl Entities {
         Ok(meta.location)
     }
 
+    /// Returns `None` if the given id would represent an index outside of `meta`.
```
Ralith, line 239:
```suggestion
    /// Returns `None` if the given id would represent an index outside of `meta`.
    ///
    /// # Safety
    /// Must only be called for currently allocated `id`s.
```

src/entities.rs
```diff
@@ -236,6 +236,22 @@ impl Entities {
...
+    pub unsafe fn resolve_unknown_gen(&self, id: u32) -> Option<Entity> {
+        let meta_len = self.meta.len();
+        if meta_len + self.pending.load(Ordering::Relaxed) as usize <= id as usize {
+            None
```
Ralith, line 243: Since such an `id` *definitely* violates the safety constraint, let's just panic here and have a simpler signature which does not give the impression of having well-defined error conditions.

src/world.rs
```diff
@@ -305,6 +305,21 @@ impl World {
...
+    ///
+    /// This function is useful if you know there's an element with a given index, but you
+    /// don't know its generation.
+    ///
+    /// # Safety
+    /// This function can still return `Some` if passed the ID of a dead entity. In this case,
+    /// it will return an entity which will, if looked up, have no components, but which may
+    /// later become allocated with some other set of components. Only use this if you're
+    /// absolutely certain that the entity is live, or if you have some other way to ensure
+    /// you have acquired a live entity!
```
Ralith, line 318:
```suggestion
    /// `id` must correspond to a currently live `Entity`. A despawned or never-allocated `id` will produce undefined behavior.
```

We do not want to guarantee any particular behavior here.

Ralith (changes requested):

src/entities.rs
sdleffler: Changed as requested; please see latest commit.

Ralith (approved):
Thanks!

src/world.rs
```diff
@@ -305,6 +305,17 @@ impl World {
...
     }
 
+    /// Given the ID of an entity without a generation, try to find the matching generation.
+    ///
+    /// This function is useful if you know there's an element with a given index, but you
+    /// don't know its generation.
+    ///
+    /// # Safety
+    /// `id` must correspond to a currently live `Entity`. A despawned or never-allocated `id` will produce undefined behavior.
+    pub unsafe fn resolve_unknown_gen(&self, id: u32) -> Entity {
```
Ralith, line 315:
```suggestion
    pub unsafe fn entity_from_id(&self, id: u32) -> Entity {
```
On further consideration, how do you feel about renaming this? We don't currently expose any user-visible notion of `generation`, so I think this is a bit more obvious (not that we want to encourage incautious use...).

sdleffler: That makes sense to me. I was struggling with finding a good name and this seems to fit, so if it satisfies you I'm down with it.

src/world.rs
```diff
@@ -305,6 +305,17 @@ impl World {
...
     }
 
+    /// Given the ID of an entity without a generation, try to find the matching generation.
+    ///
+    /// This function is useful if you know there's an element with a given index, but you
+    /// don't know its generation.
+    ///
+    /// # Safety
+    /// `id` must correspond to a currently live `Entity`. A despawned or never-allocated `id` will produce undefined behavior.
+    pub unsafe fn entity_from_id(&self, id: u32) -> Entity {
```
Ralith, line 315:
```suggestion
    /// Given an id obtained from `Entity::id`, reconstruct the still-live `Entity`.
    ///
    /// This function is useful if you know there's an element with a given index, but you
    /// don't know its generation.
    ///
    /// # Safety
    /// `id` must correspond to a currently live `Entity`. A despawned or never-allocated `id` will produce undefined behavior.
    pub unsafe fn find_entity_from_id(&self, id: u32) -> Entity {
```
Whoops, should've noticed the comment too. Also tweaking the name a little more to distinguish it from `fn entity`.

Ralith (approved):

PR #86, opened 2020-10-12
Rewrote concurrent entity allocations to be more efficient

mjhostet:
This makes several changes:

- Cuts in half the number of atomic operations to reserve an Entity,
  and simplifies the reservation code to no longer need a retry loop.
- Adds a new `reserve_entities` method that can reserve any number
  of Entity IDs with a single atomic subtract.
- Removed the `reserved` and `free` arrays, which consumed
  two u32 values for every Entity ID, and replaced them with a single
  `pending` Vec. `pending` only needs one slot for every freed Entity.
- No longer dump huge lists of fresh IDs into the freelist when we
  grow. The freelist only holds explicitly freed values, and we use
  arithmetic to acquire new IDs.
- Removes the `reserved_len()`, etc. API used to flush reserved
  entities. Now flush() takes a closure to initialize each entity.
- Tightened the restrictions on what you are allowed to do while you
  have unflushed reservations.
- Grow meta and pending incrementally, as values are allocated and
  freed, rather than in a monolithic `grow` method. This turns out
  to make things simpler, e.g. we need to track the end of the
  "reserved" range of pending anyway, and using the Vec len is handy.
- Adds some unit tests.

### DIFF
--- a/src/entities.rs
+++ b/src/entities.rs
@@ -1,7 +1,7 @@
-use alloc::boxed::Box;
 use alloc::vec::Vec;
+use core::cmp;
 use core::convert::TryFrom;
-use core::sync::atomic::{AtomicU32, Ordering};
+use core::sync::atomic::{AtomicI64, Ordering};
 use core::{fmt, mem};
 #[cfg(feature = "std")]
 use std::error::Error;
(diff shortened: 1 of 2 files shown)


mjhostet:
I assume that `contains` returning `true` for out of range IDs was a typo, and not intentionally trying to return `true` for reserved IDs. But if I'm wrong let me know.

Ralith:
Thanks for the PR! Will try to dig into it this week.

> I assume that contains returning true for out of range IDs was a typo, and not intentionally trying to return true for reserved IDs. But if I'm wrong let me know.

This was actually intended behavior. The principle is that sensible results are only guaranteed for `Entity` IDs that were legitimately obtained from the same world, and for those, the only possible way to end up with an out-of-range entity is by reserving it and not flushing the world after, therefore they're necessarily live. I'm trying to minimize the differences between a reserved entity and a conventionally spawned empty entity, at least short of showing up in queries.

src/entities.rs
```diff
@@ -52,155 +52,237 @@ impl fmt::Debug for Entity {
...
+                id,
+            };
+            Some(entity)
+        } else if let Some(id) = self.id_range.next() {
+            // The freelist ran out, so we reserved a new ID.
+            let entity = Entity { generation: 0, id };
+            Some(entity)
+        } else {
+            None
+        }
```
Ralith, line 85: Could this be more concise/elegant if `id_slice` was a `std::slice::Iter` and using `Option::or_else` rather than chained `if`s?

mjhostet: Good call, I switched `id_slice` to `id_iter`. But I don't see `or_else` or `map_or_else` working out any better here.

Also I restored the old `contains` behavior (though now using `map_or`) and added a comment and unit test.

src/world.rs
```diff
@@ -152,6 +166,8 @@ impl World {
     /// reserved entities to become real entities before proceeding. This can also be done
     /// explicitly by calling `flush`.
     ///
+    /// This is equivalent to `reserve_entities(1)`, but more efficient.
+    ///
     /// Useful for reserving an ID that will later have components attached to it with `insert`.
```
Ralith, line 171: Rather than duplicating the comment (which will certainly lead to divergence), let's direct the reader to `reserve_entities`, as is done in e.g. `insert_one` vs. `insert`.

Ralith (changes requested):
Did a deep dive. This looks awesome, thanks! Only cosmetic issues, AFAICT.

mjhostet:
Thanks, I applied all of your feedback.

src/entities.rs
```diff
@@ -366,4 +412,119 @@ mod tests {
         };
         assert_eq!(Entity::from_bits(e.to_bits()), e);
     }
+
+    #[test]
+    fn alloc_and_free() {
+        let mut rng = StdRng::seed_from_u64(0xFEEDFACEDEADF00D);
```
Ralith, line 418: That's a new one for me :grinning:

Ralith:
Great stuff!

PR #89, opened 2020-10-16
Adds spawn_at API to World.

AngelOfSol:
`spawn_at` takes an additional parameter of an Entity, to spawn the
corresponding entity at.  This can be useful if you're saving and
restoring gamestate in memory.  This can also be used to resurrect dead
entities references at will, presuming you have an old handle laying
around.

### DIFF
--- a/src/entities.rs
+++ b/src/entities.rs
@@ -224,6 +224,35 @@ impl Entities {
         }
     }
 
+    /// Allocate a specific entity ID, overwriting its generation
+    ///
+    /// Returns the location of the entity currently using the given ID, if any. Location should be written immediately.
+    pub fn alloc_at(&mut self, entity: Entity) -> Option<Location> {
+        self.verify_flushed();
+
+        let loc = if entity.id as usize >= self.meta.len() {
+            self.pending.extend((self.meta.len() as u32)..entity.id);
+            let new_free_cursor = self.pending.len() as i64;
+            self.free_cursor.store(new_free_cursor, Ordering::Relaxed); // Not racey due to &mut self
+            self.meta.resize(entity.id as usize + 1, EntityMeta::EMPTY);
+            None
+        } else if let Some(index) = self.pending.iter().position(|item| *item == entity.id) {
+            self.pending.swap_remove(index);
+            let new_free_cursor = self.pending.len() as i64;
+            self.free_cursor.store(new_free_cursor, Ordering::Relaxed); // Not racey due to &mut self
+            None
+        } else {
+            Some(mem::replace(
+                &mut self.meta[entity.id as usize].location,
+                EntityMeta::EMPTY.location,
+            ))
+        };
+
+        self.meta[entity.id as usize].generation = entity.generation;
+
+        loc
+    }
+
     /// Destroy an entity, allowing it to be reused
     ///
     /// Must not be called while reserved entities are awaiting `flush()`.
@@ -453,6 +482,50 @@ mod tests {
                 free_set.insert(id);
             }
         }
+    }
+
+    #[test]
+    fn alloc_at() {
+        let mut e = Entities::default();
+
+        let mut old = Vec::new();
+
+        for _ in 0..2 {
+            let entity = e.alloc();
+            old.push(entity);
+            e.free(entity).unwrap();
+        }
+
+        let id = old.first().unwrap().id();
+        assert!(old.iter().all(|entity| entity.id() == id));
+
+        let entity = *old.last().unwrap();
+        // The old ID shouldn't exist at this point, and should exist
+        // in the pending list.
+        assert!(!e.contains(entity));
+        assert!(e.pending.contains(&entity.id()));
+        // Allocating an entity at an unused location should not cause a location to be returned.
+        assert!(e.alloc_at(entity).is_none());
+        assert!(e.contains(entity));
+        // The entity in question should not exist in the free-list once allocated.
+        assert!(!e.pending.contains(&entity.id()));
+        // Allocating at the same id again should cause a location to be returned
+        // this time around.
+        assert!(e.alloc_at(entity).is_some());
+        assert!(e.contains(entity));
+
+        // Allocating an Entity should cause the new empty locations
+        // to be located in the free list.
+        assert_eq!(e.meta.len(), 1);
+        assert!(e
+            .alloc_at(Entity {
+                id: 3,
+                generation: 2,
+            })
+            .is_none());
+        assert_eq!(e.pending.len(), 2);
+        assert_eq!(&e.pending, &[1, 2]);
+        assert_eq!(e.meta.len(), 4);
     }
 
     #[test]
--- a/src/world.rs
+++ b/src/world.rs
@@ -36,6 +36,19 @@ use crate::{
 ///
 /// The components of entities who have the same set of component types are stored in contiguous
 /// runs, allowing for extremely fast, cache-friendly iteration.
+///
+/// There is a maximum number of unique entity IDs, which means that there is a maximum number of live
+/// entities. When old entities are despawned, their IDs will be reused on a future entity, and
+/// old `Entity` values with that ID will be invalidated.
+///
+/// While using a world, spawning and despawning billions of entities may cause a rare `Entity`
+/// value collision. A newly spawned entity after this point may have an identical complete
+/// 'Entity' value as a previously despawned entity. If an extremly long-lived `Entity` value
+/// that referred to that previously despawned entity is still around by the time this happens,
+/// that long-lived `Entity` value will erroneously refer to the newly spawned entity, rather
+/// than being considered a reference to a dead entity. This is a very rare issue as an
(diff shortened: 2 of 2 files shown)


Ralith:
Needs rebase, and I'm guessing googlebot's confused by email addresses not quite lining up.

Crazy idea: what if we named this `resurrect`?

AngelOfSol:
I thinkI think the CLA issue is because I based it off the other PR and the other commit is not author'd by me.  Resurrect is an interesting API naming choice, as it could be used as that and it's not exactly wrong per se, but on an empty world, there are no enitties to resurrect.

src/world.rs
```diff
@@ -106,6 +106,71 @@ impl World {
...
+            let index = archetype.allocate(entity.id);
+            components.put(|ptr, ty| {
+                archetype.put_dynamic(ptr, ty.id(), ty.layout().size(), index);
+            });
+            self.entities.meta[entity.id as usize].location = Location {
+                archetype: archetype_id,
+                index,
+            };
+        }
+        entity
```
Ralith, line 172: Can we share the bulk of this actual implementation with `spawn`?

Ralith (changes requested):

Ralith:
The rebase was not performed correctly.

AngelOfSol:
> The rebase was not performed correctly.

Yeah I can see that, not used to the whole PR process sorry!  I'll fix it.

src/world.rs
```diff
@@ -106,6 +106,71 @@ impl World {
...
+    /// Create an entity with certain components at a specific ID and generation.
+    ///
+    /// Returns the ID of the newly created entity.
+    ///
+    /// Arguments can be tuples, structs annotated with `#[derive(Bundle)]`, or the result of
+    /// calling `build` on an `EntityBuilder`, which is useful if the set of components isn't
+    /// statically known. To spawn an entity with only one component, use a one-element tuple like
+    /// `(x,)`.
+    ///
+    /// Any type that satisfies `Send + Sync + 'static` can be used as a component.
```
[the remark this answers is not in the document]
AngelOfSol: Done

src/world.rs
```diff
@@ -106,6 +106,71 @@ impl World {
...
+    ///
+    /// Since this reuses an existing entity id, previously dead Entity handles can become alive again,
+    /// directly (via the existing id), or indirectly (after despawning the spawn_at'd handle, and
+    /// spawning more entities).
```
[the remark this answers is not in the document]
AngelOfSol: Let me know what you think of the new wording, plus the additional paragraph at the top.

src/world.rs
AngelOfSol: I've created a spawn_inner, that de-duplicates this code.

src/entities.rs
```diff
@@ -224,6 +224,40 @@ impl Entities {
         }
     }
 
+    /// Allocate a specific entity ID and generation.  May return the location of the entity
+    /// previously allocated to the specified ID.
+    ///
+    /// Location should be written immediately.
```
Ralith, line 230:
```suggestion
    /// Allocate a specific entity ID, overwriting its generation
    ///
    /// Returns the location of the entity currently using the given ID, if any. Location should be written immediately.
```

src/entities.rs
```diff
@@ -224,6 +224,40 @@ impl Entities {
...
+    pub fn alloc_at(&mut self, entity: Entity) -> Option<Location> {
+        self.verify_flushed();
+
+        let loc = if entity.id as usize >= self.meta.len() {
+            self.pending.extend((self.meta.len() as u32)..entity.id);
```
Ralith, line 235: `self.free_cursor` must be updated whenever `self.pending` changes size, or the next `flush` will malfunction.

src/entities.rs
```diff
@@ -224,6 +224,40 @@ impl Entities {
...
+            self.meta.resize(entity.id as usize + 1, EntityMeta::EMPTY);
+            None
+        } else if let Some(index) = self
+            .pending
+            .iter()
+            .find(|item| **item == entity.id)
+            .copied()
+        {
+            self.pending.swap_remove(index as usize);
```
Ralith, line 244: `index` is an element of `pending`, not a location within it. `swap_remove` takes a location within the vec.

src/world.rs
```diff
@@ -36,6 +36,15 @@ use crate::{
 ///
 /// The components of entities who have the same set of component types are stored in contiguous
 /// runs, allowing for extremely fast, cache-friendly iteration.
+///
+/// There is a max number of unique entity IDs, which means that there is a max number of live
```
Ralith, line 40:
```suggestion
/// There is a maximum number of unique entity IDs, which means that there is a maximum number of live
```
No need to abbreviate in prose.

src/world.rs
```diff
@@ -36,6 +36,15 @@ use crate::{
...
+/// entities. When old entities are despawned, their IDs will be reused on a future entity, and
+/// old entity handles with that ID will be invalidated.
+///
+/// If an ID is reused enough times (around 4 billion times), previously dead entity handles
+/// will erroneously become live again.  The likeliness of this issue occurring is pretty rare
+/// as an ID needs to be reused enough times AND an old handle needs to be kept around in the
+/// mean time.
```
Ralith, line 47: > previously dead entity handles will erroneously become live again

This reads as if *all* previously dead handles will simultaneously become live, which is incorrect.

> The likeliness of this issue occurring is pretty rare

Could be phrased more concisely as "This is unlikely to occur" or similar.

> ID

The notion of "ID" used here and above may not be very clear to the reader, since we use that term to refer to `Entity` values elsewhere. Maybe just describe in more broad terms that after billions of entities have been despawned (specifically) it is possible for a complete `Entity` value to be reused, which may collide with any extremely long-lived references to dead entities?

src/world.rs
```diff
@@ -83,6 +92,55 @@ impl World {
...
+    /// See `spawn`.
+    ///
+    /// This will overwrite an existing entity, if the entity IDs are the same.
+    ///
+    /// Spawning with an older Entity handle, will cause other copies of the same Entity
+    /// handles to become live again.
+    ///
+    /// Be cautious with using this functionality to resurrect old entity handles as it
+    /// will vastly increase the likelihood of encountering the aforementioned issue
+    /// of dead entity handles erroneously becoming live again.
```
Ralith, line 112:
```suggestion
    /// Be cautious resurrecting old entity handles as it vastly increases
    /// the likelihood of dead entity handles erroneously becoming live again.
```

src/world.rs
```diff
@@ -83,6 +92,55 @@ impl World {
...
+        self.spawn_inner(entity, components);
+
+        entity
+    }
+
+    /// Create an entity with certain components with a specific Entity handle.
+    ///
+    /// See `spawn`.
+    ///
+    /// This will overwrite an existing entity, if the entity IDs are the same.
```
Ralith, line 105:
```suggestion
    /// Despawns any existing entity with the same `Entity::id`.
```

src/world.rs
```diff
@@ -83,6 +92,55 @@ impl World {
...
+    ///
+    /// Spawning with an older Entity handle, will cause other copies of the same Entity
+    /// handles to become live again.
```
Ralith, line 108: I'd omit this as implied by the first line.

src/world.rs
```diff
@@ -83,6 +92,55 @@ impl World {
         self.flush();
 
         let entity = self.entities.alloc();
+
+        self.spawn_inner(entity, components);
+
+        entity
+    }
+
+    /// Create an entity with certain components with a specific Entity handle.
```
Ralith, line 101:
```suggestion
    /// Create an entity with certain components and a specific Entity handle
```

src/world.rs
```diff
@@ -83,6 +92,55 @@ impl World {
...
+    ///
+    /// # Example
+    /// ```
+    /// # use hecs::*;
+    /// let mut world = World::new();
+    /// let a = world.spawn((123, "abc"));
+    /// let b = world.spawn((456, true));
+    /// world.despawn(a);
+    /// assert!(!world.contains(a));
+    /// // all previous references to 'a' will be valid again, but pointing to the new entity 'c'.
```
Ralith, line 122:
```suggestion
    /// // all previous references to 'a' will be valid again, but pointing to the new entity.
```
There is no 'c' introduced.

Ralith (changes requested):

src/world.rs
AngelOfSol: That sound good.

src/entities.rs
AngelOfSol: Should be fixed, please confirm.

src/entities.rs
AngelOfSol: Adjusted condition to use `.position` so swap_remove should work correctly here.

Ralith (approved):
This looks good; sorry it took me so long to get back to it! The contribution is much appreciated.

PR #92, opened 2020-10-18
Introduce mutation tracking query helper

Ralith:
Seeking feedback. Some initial thoughts:

- Bikeshedding welcome on the names of things
- `Modified::unset` might be annoying to invoke, particularly if you have lots of them. Is some sort of helper needed? What might that look like?

CC @cart, @sdleffler

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -69,6 +69,7 @@ mod entities;
 mod entity_builder;
 mod query;
 mod query_one;
+mod tracked;
 mod world;
 
 pub use archetype::Archetype;
@@ -78,6 +79,7 @@ pub use entities::{Entity, NoSuchEntity};
 pub use entity_builder::{BuiltEntity, EntityBuilder};
 pub use query::{Access, BatchedIter, Query, QueryBorrow, QueryIter, With, Without};
 pub use query_one::QueryOne;
+pub use tracked::{Modified, Tracked};
 pub use world::{ArchetypesGeneration, Component, ComponentError, Iter, SpawnBatchIter, World};
 
 // Unstable implementation details needed by the macros
--- a/src/tracked.rs
+++ b/src/tracked.rs
@@ -1,0 +1,131 @@
+use core::{
+    marker::PhantomData,
+    ops::{Deref, DerefMut},
+};
+
+use crate::{Access, Archetype, Component, Fetch, Query};
+
+/// Query that tracks mutable access to a component
+///
+/// Using this in a query is equivalent to `(&mut T, &mut Modified<T>)`, except that it yields a
+/// smart pointer to `T` which sets the flag inside `Modified<T>` to `true` when it's mutably
+/// borrowed.
+///
+/// A `Modified<T>` component must exist on an entity for it to be exposed to this query.
+///
+/// # Example
+/// ```
+/// # use hecs::*;
+/// let mut world = World::new();
+/// let e = world.spawn((123, Modified::<i32>::new()));
+/// for (_id, mut value) in world.query::<Tracked<i32>>().iter() {
+///   assert_eq!(*value, 123);
+/// }
+/// assert!(!world.get::<Modified<i32>>(e).unwrap().is_set());
+/// for (_id, mut value) in world.query::<Tracked<i32>>().iter() {
+///   *value = 42;
+/// }
+/// assert!(world.get::<Modified<i32>>(e).unwrap().is_set());
+/// ```
+pub struct Tracked<'a, T: Component> {
+    value: &'a mut T,
+    modified: &'a mut Modified<T>,
+}
+
+impl<'a, T: Component> Deref for Tracked<'a, T> {
+    type Target = T;
+    #[inline]
+    fn deref(&self) -> &T {
+        self.value
+    }
+}
+
+impl<'a, T: Component> DerefMut for Tracked<'a, T> {
+    #[inline]
+    fn deref_mut(&mut self) -> &mut T {
+        self.modified.0 = true;
+        self.value
+    }
+}
+
+impl<'a, T: Component> Query for Tracked<'a, T> {
+    type Fetch = FetchTracked<T>;
+}
+
+/// A flag indicating whether the `T` component was modified
+///
+/// Must be manually added to components that will be queried with `Tracked`.
+pub struct Modified<T>(bool, PhantomData<T>);
+
+impl<T> Modified<T> {
+    /// Constructs an unset flag
+    #[inline]
+    pub fn new() -> Self {
+        Self(false, PhantomData)
+    }
+
+    /// Returns whether the `T` component was modified since the last `unset` call
+    #[inline]
+    pub fn is_set(&self) -> bool {
+        self.0
+    }
+
+    /// Unsets the flag
+    #[inline]
+    pub fn unset(&mut self) {
+        self.0 = false;
+    }
+}
+
+impl<T> Default for Modified<T> {
+    #[inline]
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[doc(hidden)]
+pub struct FetchTracked<T: Component> {
+    value: <&'static mut T as Query>::Fetch,
+    modified: <&'static mut Modified<T> as Query>::Fetch,
+}
+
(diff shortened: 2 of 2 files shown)


AngelOfSol:
To be clear on the user side, if you don't use a query with `Tracked<T>` the change will go untracked?  Does this mean you can only query either: all instances of component T with no tracking or just `Tracked<T>` with tracking?

Ralith:
Correct. You might legitimately want to exempt a system from mutation tracking, even.

Ratysz:
Bevy has query transformers that yield only the mutated items, which I think is the only reason for a mechanism more involved than a mundane `dirty: bool` field on user side. (In Bevy that's built on `Fetch::should_skip()`.)

Side note, this transformer approach to tracking is quite elegant. Might it be warranted to officially expose `Fetch`, so that users could implement their own convenience transformers?

> Is some sort of helper needed? What might that look like?

Right now resetting tracking for all components of type `T` looks like this (is that correct?):

```rust
for (_, mut modified) in world.query::<Modified<T>>().iter() {
    modified.unset();
}
```

Perhaps something as simple as `World::reset_tracking_for_all<T: Component>(&self)`, doing just that?

Additionally, `unset()` method(s) on `Tracking` itself should allow solving the situation of "do something with component and reset it's tracking" in one query, rather than two. Alternatively, another specialized transformer just for that case.

Combining this with filtering queries from the first point creates a "drain" iterator for mutated components.

Ralith:
> query transformers that yield only the mutated items .. built on `Fetch::should_skip()`

I feel like this is a bit redundant to `filter`ing the iterator. Is the interface significantly more convenient?

> Might it be warranted to officially expose Fetch, so that users could implement their own convenience transformers?

I think this is a great extension point in principle. I'd be much happier if it wasn't so unsafe, though. I wonder if there's some way we can expose composition primitives without requiring people to implement all the hazardous methods.

> Perhaps something as simple as World::reset_tracking_for_all<T: Component>(&self), doing just that?

I'm not sure we need a oneliner to replace a threeliner. I'm more concerned about bevy's case, where there's a large number of modification states to reset (one per component). If they're currently relying on hard-coded pseudo-components, they might be having an easier time iterating over and clearing them than this interface permits, particularly given that the set of components bevy wants to reset tracking state for isn't statically known.

> do something with component and reset it's tracking

Would it be sufficient to just query the `T` and `Modified<T>` separately for that case?

Ratysz:
> Is the interface significantly more convenient?

Tentatively, yes, if only because it's more conscise. However, Bevy does expose a lot of transformers for that, to the point where I don't know which one I want without looking at docs. I suppose examples of this filtering in docs would be a better solution than exposing more types.

> I'm more concerned about bevy's case, where there's a large number of modification states to reset (one per component).

Right; I haven't had an idea about that yet.

> Would it be sufficient to just query the `T` and `Modified<T>` separately for that case?

Honestly, that didn't occur to me, somehow in this instance I forgot that those two are actually orthogonal components. Yes, it is sufficient (with doc examples).

cart:
I think this approach is probably the right call for hecs because it (1) lets users opt-in to tracking without touching the default hecs code path at all (2) very straightforward.

It also _might_ be a good call for bevy. Functionally its basically the same, but there are a few UX issues I'd probably want to solve (either in hecs upstream or in a bevy fork). First, I'm a little bit hesitant to make users filter like this:

```rust
// this pr
for mut modified in world.query::<&mut Modified<T>>().iter().filter(|m| m.is_set()) {
}

// bevy
for mut modified in world.query::<&mut Modified<T>>().iter() {
}
```

If Bevy continues with the "universal" track-everything approach, it would probably mean we'd need to fork to prevent people from "breaking out" of the tracking. Ex: un-impling Query for `&mut T`. (which is what we currently do).

Additionally, resetting the trackers using this pr requires somehow composing a strongly-typed "reset system" for each component. Automatically resetting all component trackers on behalf of users presents a bit of a challenge.

And as mentioned in the other pr, auto-inserting the tracking components on behalf of users is also a challenge (but probably a solveable one).

I'd also like to point out that tracking "added" components doesn't fit (directly) into this model, which is something Bevy would need.

Ralith:
> First, I'm a little bit hesitant to make users filter like this:

Yeah, that does kind of suck, especially for more complex queries where you have to pick out the right field. I don't think incorporating linear filtering into queries incurs any cost or significant complexity, so ergonomics is sufficient argument to support it.

> we'd need to fork to prevent people from "breaking out" of the tracking

> auto-inserting the tracking components on behalf of users

I suspect this can be accomplished with some sort of wrapper, though perhaps tediously. A bevy-owned newtype of `World` could then apply the wrapper internally to the interfaces it chooses to expose, without duplicating the heavy lifting/unsafe stuff.

> Automatically resetting all component trackers on behalf of users presents a bit of a challenge.

There's a pattern that comes up a lot which we could use here: accumulate a `Vec<Box<dyn Fn(&World)>>` with an entry for every unique component. It's not super graceful and it's (very fractionally) less efficient than iterating over archetypes first, though. I wonder if there's some way we could instead have one or more `Box<dyn Fn>`s per archetype with user-controlled per-component code inside...

> tracking "added" components doesn't fit (directly) into this model

What are the semantics of that, exactly? Can a component be deemed both added and modified, or are they mutually exclusive?

One interesting trick might be to store, for each entity, the archetype it was last in. You could infer added and removed components freely by diffing the component lists against the current one.

Ralith:
(also, as much as it's a juicy extension point, I'm glad we haven't exposed `Fetch` just yet because #95 and #96 just brought us a *spectacular* speedup by vigorously breaking it)

cart:
Well thats _really_ cool. I'm curious to see if those wins will apply to `bevy_hecs` too, or if the changes I made result in similar assembly as the PRs above. Compiler optimizations are random enough that it always just feels like magic to me.

Lets hope these wins stack :)

cart:
> Yeah, that does kind of suck, especially for more complex queries where you have to pick out the right field. I don't think incorporating linear filtering into queries incurs any cost or significant complexity, so ergonomics is sufficient argument to support it.

Yeah I think its a nice ergo win. But I do want to point out that the "linear filter" breaks things like cheap iterator length via ExactSizeIterator.

> I suspect this can be accomplished with some sort of wrapper, though perhaps tediously. A bevy-owned newtype of World could then apply the wrapper internally to the interfaces it chooses to expose, without duplicating the heavy lifting/unsafe stuff.

Yeah that could definitely be possible. Regardless, I don't think its fair to constrain hecs' design according to bevy's constraints. I do want us to consume upstream directly if we can (for the benefit of both projects), but I also don't want to force hecs in a direction thats suboptimal for non-bevy scenarios.

> There's a pattern that comes up a lot which we could use here: accumulate a Vec<Box<dyn Fn(&World)>> with an entry for every unique component. It's not super graceful and it's (very fractionally) less efficient than iterating over archetypes first, though. I wonder if there's some way we could instead have one or more Box<dyn Fn>s per archetype with user-controlled per-component code inside...

Cool yeah that would would / seems acceptable to me. User-configurable per-archetype/component code is also an interesting prospect.

> What are the semantics of that, exactly? Can a component be deemed both added and modified, or are they mutually exclusive?

We store "added" and "mutated"  flags for each component. "Added" is set when the component is first inserted.  "Mutated" is set only when a user mutates the component. We then have Added, Mutated, and Changed queries/pointers where Added and Mutated correspond to their component states and Changed is `added | mutated`.

Mutated vs Changed is a little confusing, but it lets us capture every relevant scenario.

I've thought a lot about removed states (which we also support). Right now we just store a list of removed entities for each component and reset it each frame. But I really want us to be able to query removed components. I think this is possible because part of the removal process is swapping the removed component to the end of the "array". As long as that space isn't overwritten by a new component, the memory there is still valid. With a little bit of extra state management we could ensure that space isn't overwritten until a frame later.

zakarumych:
User should be able to convert Mut<'a, T> into &'a mut T, which is now impossible as there's only DerefMut which returns mutable reference with lifetime limited by Mut value itself.

Ralith:
Closing in favor of https://github.com/Ralith/hecs/pull/366, which is more ergonomic and should perform better in most cases.

PR #93, opened 2020-10-18
Use a much faster hash for TypeId

mjhostet:
TypeId is already a fully-hashed u64, so there is no need to hash it further.
Instead, just use nohash-hasher to pass it through unchanged.

### DIFF
--- a/src/archetype.rs
+++ b/src/archetype.rs
@@ -17,10 +17,11 @@ use crate::alloc::boxed::Box;
 use crate::alloc::{vec, vec::Vec};
 use core::any::{type_name, TypeId};
 use core::cell::UnsafeCell;
+use core::hash::{BuildHasher, BuildHasherDefault, Hasher};
 use core::mem;
 use core::ptr::{self, NonNull};
 
-use hashbrown::HashMap;
+use hashbrown::{hash_map::DefaultHashBuilder, HashMap};
 
 use crate::borrow::AtomicBorrow;
 use crate::query::Fetch;
@@ -32,7 +33,7 @@ use crate::{Access, Component, Query};
 /// go through the `World`.
 pub struct Archetype {
     types: Vec<TypeInfo>,
-    state: HashMap<TypeId, TypeState>,
+    state: TypeIdMap<TypeState>,
     len: u32,
     entities: Box<[u32]>,
     // UnsafeCell allows unique references into `data` to be constructed while shared references
@@ -198,7 +199,7 @@ impl Archetype {
             self.entities = new_entities;
 
             let old_data_size = mem::replace(&mut self.data_size, 0);
-            let mut state = HashMap::with_capacity(self.types.len());
+            let mut state = HashMap::with_capacity_and_hasher(self.types.len(), Default::default());
             for ty in &self.types {
                 self.data_size = align(self.data_size, ty.layout.align());
                 state.insert(ty.id, TypeState::new(self.data_size));
@@ -337,6 +338,50 @@ impl Drop for Archetype {
         }
     }
 }
+
+/// A hasher optimized for hashing a single TypeId.
+///
+/// TypeId is already thoroughly hashed, so there's no reason to hash it again.
+/// Just leave the bits unchanged.
+#[derive(Default)]
+pub(crate) struct TypeIdHasher {
+    hash: u64,
+}
+
+impl Hasher for TypeIdHasher {
+    fn write_u64(&mut self, n: u64) {
+        // Only a single value can be hashed, so the old hash should be zero.
+        debug_assert_eq!(self.hash, 0);
+        self.hash = n;
+    }
+
+    // Tolerate TypeId being either u64 or u128.
+    fn write_u128(&mut self, n: u128) {
+        debug_assert_eq!(self.hash, 0);
+        self.hash = n as u64;
+    }
+
+    fn write(&mut self, bytes: &[u8]) {
+        debug_assert_eq!(self.hash, 0);
+
+        // This will only be called if TypeId is neither u64 nor u128, which is not anticipated.
+        // In that case we'll just fall back to using a different hash implementation.
+        let mut hasher = DefaultHashBuilder::new().build_hasher();
+        hasher.write(bytes);
+        self.hash = hasher.finish();
+    }
+
+    fn finish(&self) -> u64 {
+        self.hash
+    }
+}
+
+/// A HashMap with TypeId keys
+///
+/// Because TypeId is already a fully-hashed u64 (including data in the high seven bits,
+/// which hashbrown needs), there is no need to hash it again. Instead, this uses the much
+/// faster no-op hash.
+pub(crate) type TypeIdMap<V> = HashMap<TypeId, V, BuildHasherDefault<TypeIdHasher>>;
 
 struct TypeState {
     offset: usize,
--- a/src/entity_builder.rs
+++ b/src/entity_builder.rs
@@ -19,9 +19,9 @@ use core::any::TypeId;
 use core::mem::{self, MaybeUninit};
 use core::ptr;
 
-use hashbrown::hash_map::{Entry, HashMap};
+use hashbrown::hash_map::Entry;
 
-use crate::archetype::TypeInfo;
+use crate::archetype::{TypeIdMap, TypeInfo};
 use crate::{Component, DynamicBundle};
 
 /// Helper for incrementally constructing a bundle of components with dynamic component types
@@ -42,7 +42,7 @@ pub struct EntityBuilder {
     cursor: usize,
     info: Vec<(TypeInfo, usize)>,
     ids: Vec<TypeId>,
-    indices: HashMap<TypeId, usize>,
+    indices: TypeIdMap<usize>,
 }
 
 impl EntityBuilder {
@@ -53,7 +53,7 @@ impl EntityBuilder {
             storage: Box::new([]),
             info: Vec::new(),
             ids: Vec::new(),
-            indices: HashMap::new(),
+            indices: Default::default(),
         }
     }
 


Cargo.toml
```diff
@@ -26,6 +26,7 @@ macros = ["hecs-macros", "lazy_static"]
 hecs-macros = { path = "macros", version = "0.3.0", optional = true }
 hashbrown = { version = "0.9.0", default-features = false, features = ["ahash", "inline-more"] }
 lazy_static = { version = "1.4.0", optional = true, features = ["spin_no_std"] }
+nohash-hasher = { version = "0.2.0" }
```
Ralith, line 29:
```suggestion
nohash-hasher = { version = "0.2.0", default-features = false }
```
for no-std support

Ralith (changes requested):
Nice!

Ralith:
Out of curiosity, does this register on benches?

Cargo.toml
mjhostet: Also, if you don't like pulling in this crate, it's easy to do this inline; here's where I tweaked legion's version of this: https://github.com/amethyst/legion/pull/204

mjhostet:
This general trick is a measurable win in a compiler I work on, but I wasn't sure how to benchmark here. I tried "cargo bench" and it just prints a bunch of "ignored" lines but I didn't look into it.

src/archetype.rs
```diff
@@ -26,13 +26,20 @@ use crate::borrow::AtomicBorrow;
 use crate::query::Fetch;
 use crate::{Access, Component, Query};
 
+/// A HashMap with TypeId keys
+///
+/// Because TypeId is already a fully-hashed u64 (including data in the high seven bits,
+/// which hashbrown needs), there is no need to hash it again. Instead, this uses the much
+/// faster no-op hash.
+pub type TypeIdMap<V> = HashMap<TypeId, V, nohash_hasher::BuildNoHashHasher<u64>>;
```
mjhostet, line 34: I was wondering if you would want to export this type so that someone like @cart could use it. bevy has various `HashMap<TypeId>` that could benefit, although it also `HashSet<TypeId>`, which would be a bit weird to export from hecs since you don't use it.

If you don't want to export this I can explicitly make it `pub(crate)` to be more clear.

Ralith: I'm disinclined to export it since 1. it's trivial and 2. it's out of scope for hecs proper. In principle this is a case for a shared crate, though, it's so trivial that I dunno if it's justified.

Cargo.toml
Ralith: I don't have an objection in principle to a small dep, but reviewing nohash-hasher's impl I'm a bit confused by it: the type parameter seems to be completely meaningless. Am I missing something there? Considering the strangeness there, and the benefit of ensuring that the actual no-op hash implementation gets inlined, having our own impl might be better.

Ralith:
> I tried "cargo bench" and it just prints a bunch of "ignored" lines but I didn't look into it.

`cargo bench --all-features` is needed as the benchmarks use the macros. They're not *super* comprehensive since I've just been filling them out as needed to answer specific performance questions.

cart:
Something to point out: this might not work forever as rustc devs are considering making TypeId a u128 instead of a u64 internally (which nohash doesn't support):

https://github.com/rust-lang/rust/pull/75923

Ralith:
Good catch, thanks! We should probably roll our own which tolerates arbitrary-length inputs, then. Presuming the statistical properties are similar we should be able to just truncate.

mjhostet:
This is about 33% faster on the spawn_batch benchmark (34147 ns -> 25611 ns), but strangely 11% slower on iterate_100k (179849 ns -> 202057 ns). I don't have an explanation.

src/archetype.rs
```diff
@@ -338,6 +339,44 @@ impl Drop for Archetype {
...
+    }
+
+    // Tolerate TypeId being either u64 or u128.
+    fn write_u128(&mut self, n: u128) {
+        debug_assert_eq!(self.hash, 0);
+        self.hash = n as u64;
+    }
+
+    fn write(&mut self, _bytes: &[u8]) {
+        panic!("`write` should not be called when hashing a TypeId");
```
Ralith, line 365: Because `write` is the fallback impl, to reduce the odds of existing code breaking on unforseen future rust versions, I'd prefer to do something conservative here, e.g. take the first 8 bytes, zero-extending.

Ralith (changes requested):
I confirm a ~30% improvement on spawn_batch. On my system (i7-8550U, poorly isolated) I actually see a **40% slowdown** on iterate_100k (~220ns -> ~300ns). Adding `#[inline]` annotations had no clear effect. Extremely surprising. I'd like to better understand what's going on here. Maybe we're somehow defeating hashbrown's SIMD stuff?

Ralith:
Splitting the change-of-archetype case of `QueryIter::next` into its own function narrows the gap some with this change, but makes things worse on master, bringing both to ~250μs.

cart:
Yeah given that iteration speed is the "strong point" of an archetypal ecs, I think we should optimize for that over spawn speed.

Imberflur:
On my system (i5-4690) I see
spawn_batch: ~45us-> ~32us
iterate_100k: ~207us -> ~208us

Ralith:
The outlining change I described above makes this change faster than master on today's nightly (but not on rustc 1.47), even though it still makes master slower:

```rust
diff --git a/src/query.rs b/src/query.rs
index e635e33..8c9d80f 100644
--- a/src/query.rs
+++ b/src/query.rs
@@ -432,6 +432,21 @@ pub struct QueryIter<'q, 'w, Q: Query> {
     iter: Option<ChunkIter<Q>>,
 }
 
+impl<'q, 'w, Q: Query> QueryIter<'q, 'w, Q> {
+    fn next_archetype(&mut self) -> Option<()> {
+        let archetype = self.borrow.archetypes.get(self.archetype_index as usize)?;
+        self.archetype_index += 1;
+        unsafe {
+            self.iter = Q::Fetch::get(archetype, 0).map(|fetch| ChunkIter {
+                entities: archetype.entities(),
+                fetch,
+                len: archetype.len(),
+            });
+        }
+        Some(())
+    }
+}
+
 unsafe impl<'q, 'w, Q: Query> Send for QueryIter<'q, 'w, Q> {}
 unsafe impl<'q, 'w, Q: Query> Sync for QueryIter<'q, 'w, Q> {}
 
@@ -442,17 +457,7 @@ impl<'q, 'w, Q: Query> Iterator for QueryIter<'q, 'w, Q> {
     fn next(&mut self) -> Option<Self::Item> {
         loop {
             match self.iter {
-                None => {
-                    let archetype = self.borrow.archetypes.get(self.archetype_index as usize)?;
-                    self.archetype_index += 1;
-                    unsafe {
-                        self.iter = Q::Fetch::get(archetype, 0).map(|fetch| ChunkIter {
-                            entities: archetype.entities(),
-                            fetch,
-                            len: archetype.len(),
-                        });
-                    }
-                }
+                None => self.next_archetype()?,
                 Some(ref mut iter) => match unsafe { iter.next() } {
                     None => {
                         self.iter = None;
```

I feel like something about `QueryIter::next` is really throwing llvm for a loop. Maybe @cart's inlining changes would have an impact here?

Ralith:
On advice from @FeepingCreature I got rid of the outer branch in `QueryIter::next`, ~doubling performance and annihilating the confusing discrepancies here. See https://github.com/Ralith/hecs/pull/95.

src/archetype.rs
mjhostet: I added the requested paranoia code to handle TypeId being neither u64 nor u128.

Ralith (approved):

Ralith:
Thanks!

src/archetype.rs
```diff
@@ -338,6 +339,50 @@ impl Drop for Archetype {
...
+        debug_assert_eq!(self.hash, 0);
+        self.hash = n as u64;
+    }
+
+    fn write(&mut self, bytes: &[u8]) {
+        debug_assert_eq!(self.hash, 0);
+
+        // This will only be called if TypeId is neither u64 nor u128, which is not anticipated.
+        // In that case we'll just fall back to using a different hash implementation.
+        let mut hasher = DefaultHashBuilder::new().build_hasher();
```
mjhostet, line 369: This could also just use ahash explicitly.

Ralith: I'm fine with deferring that decision upstream, even if it wasn't dead code.

cart:
Just ported #95 and #96 to bevy_hecs. As I sort of suspected, it had zero effect (maybe a slighty negative effect) on simple iteration performance, buuuut it did give some nice boosts to other benches like "fragmented iteration":

image

For reference: on my machine hecs went from ~1800ns to 990ns on simple_iter. I assume the 200ns discrepancy between hecs and bevy is the overhead of boxed systems / query setup / Fetch::should_skip.

I'm gonna go ahead and call this a huge win for hecs and a medium-sized win for bevy :)

cart:
Ah wait we aren't using bevy's System abstraction in those benches. So thats not where the overhead comes from.

cart:
Ooh I just ported the TypeId hasher and thats also a huge win for bevy:

image

Ralith:
Interesting! Can you try running hecs' benchmarks before/after #100? If you see no change there as well I'll write it off to a quirk of my environment (??) and keep the change.

cart:
Sure thing!

Ralith:
> Fetch::should_skip

Provided there's sufficient inlining (which should be achievable by annotations if it's not happening on its own) I think this should be zero-cost in the sense that queries which don't use it will have the branches compiled out.

cart:
This is the result of running the benches on the "revert" branch first, then running them again on master, so the "improvements" are _with_ "Use a much faster hash for TypeId"

image


Based on these results, I don't think reverting is the right call (at least for these benchmarks)

PR #100, opened 2020-10-21
Revert "Use a much faster hash for TypeId"

Ralith:
This reverts commits 5ffba7fc3a63e988e5966481b5842055c44f6801 and 0b78f1453d4a51fbc1ce96f256403bca9242f0a6.

These still, somehow, seem to slow down iteration significantly (~36%).

cc @mjhostet

On reexamination it looks like I was a bit hasty in declaring the strange behavior solved; this revert reduces spawn_batch perf by 25% but improves iterate_100k perf more. I'm still mystified by what's going on here. Some inlining heuristic screwup?


Ralith:
Closing per discussion in #93. This must be environmental, somehow.

PR #101, opened 2020-10-24
Use `usize` rather than `u32` in QueryIter to avoid zero extensions

mjhostet:
Loops using QueryIter spend time zero-extending u32 values to usize
so they can be used in address computations. There's no need to save
memory for iterator fields, since iterators are transient, so instead
just widen them to usize, which is more natural.

This drops the inner loop for the iterate_100k benchmark from this:

```
loop:
       mov    %esi,%eax
       movss  (%r14,%rax,4),%xmm0
       addss  0x0(%rbp,%rax,4),%xmm0
       add    $0x1,%esi
       movss  %xmm0,0x0(%rbp,%rax,4)
       mov    0x30(%rsp),%rax
       mov    %rax,(%rsp)
       mov    0x38(%rsp),%r13
       cmp    %ecx,%esi
       je     done
       test   %rbp,%rbp
       jne    loop
done:
```

to this:

```
loop:
       movss  (%r14,%rsi,4),%xmm0
       addss  0x0(%rbp,%rsi,4),%xmm0
       movss  %xmm0,0x0(%rbp,%rsi,4)
       add    $0x1,%rsi
       test   %rbp,%rbp
       je     done
       cmp    %rcx,%rsi
       jne    loop
done:
```

Resulting in a 29% speedup.

Testing if `%rbp` (one of the array pointers) is nonnull every time through
the loop still needs to be fixed -- then maybe it will vectorize.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -431,7 +431,7 @@ impl<'q, 'w, Q: Query> IntoIterator for &'q mut QueryBorrow<'w, Q> {
 /// Iterator over the set of entities with the components in `Q`
 pub struct QueryIter<'q, 'w, Q: Query> {
     borrow: &'q mut QueryBorrow<'w, Q>,
-    archetype_index: u32,
+    archetype_index: usize,
     iter: ChunkIter<Q>,
 }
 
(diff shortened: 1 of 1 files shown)


Ralith (approved):
That's awesome! I had no idea that might have such an impact.

FeepingCreature:
This seems more like a weird compiler issue. If I understand my x86 correctly, the 32-bit version should just increment and compare `%esi` instead of `%rsi`. But hey, always better to work in native word size if you're confident you're fitting in the registers.

mjhostet:
Agreed, this is really spoonfeeding the compiler. The `test %rbp,rbp` -- checking to see if the pointer it just dereferenced is null -- is especially disturbing, and still there after this change.

Ralith:
Do you know what array that comes from to begin with? I'm not sure why such
a check would exist at all.

mjhostet:
I'm not sure, but the null test seems to originate from checking whether an Option is None by comparing one of its references against the sentinel null value. But LLVM gets deeply confused and somehow keeps checking this even inside the inner loop. I attempted to unconfuse it in https://github.com/Ralith/hecs/pull/103

PR #102, opened 2020-10-24
Mark a couple archetype methods as `inline`

mjhostet:
ChunkIter calls `archetype.entities()` and `archetype.len()`.
By marking them `inline`, the outer loop of query iteration
avoids two function calls to these trivial methods (and the
register spills those cause).

This speeds up the iterate_100k benchmark by 23%. That feels
unreasonably high since it doesn't affect the innermost loop, but
even if this benefit is illusory the machine code is tighter so
this is a win.

### DIFF
--- a/src/archetype.rs
+++ b/src/archetype.rs
@@ -122,10 +122,12 @@ impl Archetype {
         }
     }
 
+    #[inline]
     pub(crate) fn len(&self) -> u32 {
         self.len
     }
 
+    #[inline]
     pub(crate) fn entities(&self) -> NonNull<u32> {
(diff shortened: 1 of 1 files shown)


mjhostet:
Even though this shouldn't be helping iterate_100k very much, it is theoretically useful where there is high archetype fragmentation, because these archetype methods will get called much more often.

Ralith (approved):
Nice!

Ralith:
I actually don't see any change in `iterate_100k` here, which makes sense, but IMO this change is justified regardless.

PR #103, opened 2020-10-25
Use functions rather than consts for DANGLING values

mjhostet:
LLVM seems to get confused by the use of consts for the `DANGLING`
structs; it's hard to see what's going on but perhaps it's losing track of
the nonnull properties of some fields, even though they are constants?
In any case, it handles equivalent functions just fine, so use those instead.
There is naming precedent with `NonNull::dangling()` anyway.

This drops the inner loop for iterate_100k from this (note the bizarre test
of the base pointer we just dereferenced against null):

```
loop:
        movss  (%rsi,%rbp,4),%xmm0
        addss  (%rcx,%rbp,4),%xmm0
        movss  %xmm0,(%rcx,%rbp,4)
        add    $0x1,%rbp
        test   %rcx,%rcx
        je     done
        cmp    %rdi,%rbp
        jne    loop
```

to this:

```
loop:
        movss  (%rsi,%rax,4),%xmm0
        addss  (%r12,%rax,4),%xmm0
        movss  %xmm0,(%r12,%rax,4)
        add    $0x1,%rax
        cmp    %rdi,%rax
        jne    loop
```

This is about an 8% speedup, although there is some noise.

It's still not vectorizing. That may be because the relevant llvm memory
operations don't have `noalias` markers for some reason, even though
one is `&mut`. The LLVM IR for the inner loop is still more complicated
than I would expect, which could also be confusing it.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -31,8 +31,7 @@ pub trait Fetch<'a>: Sized {
     type Item;
 
     /// A value on which `get` may never be called
-    #[allow(clippy::declare_interior_mutable_const)] // no const fn in traits
-    const DANGLING: Self;
+    fn dangling() -> Self;
 
     /// How this query will access `archetype`, if at all
     fn access(archetype: &Archetype) -> Option<Access>;
@@ -75,7 +74,9 @@ pub struct FetchRead<T>(NonNull<T>);
 impl<'a, T: Component> Fetch<'a> for FetchRead<T> {
     type Item = &'a T;
 
-    const DANGLING: Self = Self(NonNull::dangling());
+    fn dangling() -> Self {
+        Self(NonNull::dangling())
+    }
 
     fn access(archetype: &Archetype) -> Option<Access> {
         if archetype.has::<T>() {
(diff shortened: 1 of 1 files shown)


mjhostet:
One possible theory about recent massive benchmarking noise:

rustc partitions a unit's LLVM IR into multiple shards so they can be compiled in parallel, then optimizes them together using "thin lto". It may be arbitrary whether one of these DANGLING `const`s appears in a different shard from its use (and I definitely saw this happen). When that happens, the use cannot get optimized as much, because LLVM does not "know" the constant field values -- that llvm module only has an "external" declaration for it. I haven't looked into it at all, but maybe thin lto is better at optimizing w/external functions than const structs, so it can't "clean them up", which would be one reason this diff helps.

If noise happens again, `RUSTFLAGS="-C codegen-units=1"` (to disable sharding) would be an interesting experiment.

Ralith (approved):
Strange. Stranger still, `RUSTFLAGS="-C codegen-units=1"` seems to ~halve performance on master.

I believe noalias is omitted because it's broken in llvm: https://github.com/rust-lang/rust/issues/54878.

Ralith:
Noticable performance improvement from `codegen-units=1` after merging this, though!

Ralith:
No noticeable benefit from `-Z mutable-noalias=yes`.

PR #104, opened 2020-10-28
Rewrite and test the Bundle derive macro

Veykril:
This PR adds tests for the derive macro, makes it hygienic and fixes it not working for tuple structs and unit structs(though I guess unit structs are kind of pointless to use?).

I wanted to tackle adding generics but figured adding proper testing first would make more sense.


macros/src/lib.rs
```diff
@@ -64,54 +64,72 @@ pub fn derive_bundle(input: TokenStream) -> TokenStream {
...
+                    mut f: impl FnMut(::hecs::TypeInfo) -> Option<std::ptr::NonNull<u8>>,
+                ) -> Result<Self, ::hecs::MissingComponent> {
+                    #(
+                        let #fields = f(::hecs::TypeInfo::of::<#tys>())
+                                .ok_or_else(::hecs::MissingComponent::new::<#tys>)?
+                                .cast::<#tys>()
+                            .as_ptr();
+                    )*
+                    Ok(Self { #( #fields: #fields.read(), )* })
+                }
```
Ralith: Rather than having two code paths here, could we resolve the issue with type annotations?

tests/derive.rs
```diff
@@ -1,8 +1,8 @@
-const TEST_DIR: &str = "tests/derive";
-
 #[cfg(feature = "macros")]
 #[test]
 fn derive() {
+    const TEST_DIR: &str = "tests/derive";
+
```
Ralith, line 8: This probably belongs in the prior commit.

macros/src/lib.rs
```diff
@@ -124,35 +126,36 @@ fn gen_bundle_impl(ident: &syn::Ident, fields: &[syn::Ident], tys: &[&syn::Type]
...
+                            .as_ptr();
                 )*
-                Ok(Self { #( #fields: #fields.read(), )* })
+                ::std::result::Result::Ok(Self { #( #fields: #fields.read(), )* })
             }
         }
     }
 }
 
+// no reason to generate a static for unit structs
```
Ralith, line 157: Probably belongs in previous commit.

tests/derive.rs
```diff
@@ -1,6 +1,7 @@
 //!
 
 #[cfg(feature = "macros")]
+#[cfg_attr(miri, ignore)]
```
Ralith, line 4: Might as well fold this into the original commit.

Ralith (approved):
Thanks, this is great! No serious issues, though if you want to rewrite the history a bit we could rebase rather than squashing and keep history a little more discrete.

> though I guess unit structs are kind of pointless to use?

I could imagine them coming up occasionally with feature-gated or macro-generated component fields. Definitely worth supporting, given that it's not hugely difficult to do so.

Veykril:
Alright, the history should be cleaner now

Veykril:
Ah dammit i pushed the generics test file by accident now

PR #105, opened 2020-10-29
Support generics in bundle derive macro

### LINKED ISSUE #87: Add support for generic types to `#[derive(Bundle)]`
At current, the `Bundle` derive macro doesn't support generics and it seems to me like it should be able to support them without too much trouble.

Veykril:
Fixes #87

This still does the duplication check, unlike tuples which don't do that. Should that stay or not?

Edit: Maybe the dupe check can be made cheaper by making it lazy and only execute once via AtomicBool and closures?

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -104,13 +104,6 @@ fn gen_bundle_impl(
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
     let with_static_ids_inner = quote! {
         {
-            let mut dedup = ::std::collections::HashSet::new();
-            for &(ty, name) in [#((::std::any::TypeId::of::<#tys>(), ::std::any::type_name::<#tys>())),*].iter() {
-                if !dedup.insert(ty) {
-                    ::std::panic!("{} has multiple {} fields; each type must occur at most once!", stringify!(#ident), name);
-                }
-            }
-
             let mut tys = [#((::std::mem::align_of::<#tys>(), ::std::any::TypeId::of::<#tys>())),*];
             tys.sort_unstable_by(|x, y| {
                 ::std::cmp::Ord::cmp(&x.0, &y.0)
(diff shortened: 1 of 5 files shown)


Ralith:
We definitely don't want to be constantly rerunning the duplicate check. I think an `AtomicBool` scheme will have the same problems that a static array did, i.e. we can't have one per type parameter set. However, I think there's an easy solution here: add `assert!(types.windows(2).all(|x| x[0] != x[1]), "duplicate types");` or similar to `Archetype::new`. This will catch that case for tuples and derived bundles both without significant cost.

Veykril:
That will only catch duplicate types that are right beside each other though, as in `(i32, i64, i32)` won't fail.

I pushed a commit in which I experimented a bit, let me know what you think about it. I do check for duplicates in the Archetype constructor now and in debug builds it also emits the names of the component types in case it fails.

Theoretically this could also be made to output the type name of `Bundle`s but I didn't want to make too big changes at once now.

Ralith:
> That will only catch duplicate types that are right beside each other though

The `Vec<TypeInfo>` passed to `Archetype::new` is required to be sorted, per the debug_assert. In a sorted list, equal elements are guaranteed to be adjacent.

> in debug builds it also emits the names of the component types in case it fails.

Nice! I'd prefer to just print the name of the duplicated type rather than all the types, though. Additionally, we should find a way to communicate the error without mentioning archetypes, as the average user shouldn't be required to know what those are. Maybe "attempted to allocate an entity with duplicate components"?

Veykril:
> The Vec<TypeInfo> passed to Archetype::new is required to be sorted, per the debug_assert. In a sorted list, equal elements are guaranteed to be adjacent.
Right, not sure why I thought that wouldn't work 😅

Alright, it should print the type instead now together with your proposed message

src/archetype.rs
```diff
@@ -43,11 +43,21 @@ pub struct Archetype {
...
+    pub(crate) fn empty() -> Self {
+        Self::new_(Vec::new())
+    }
+
+    pub(crate) fn new<B: crate::DynamicBundle>(types: Vec<TypeInfo>) -> Self {
         debug_assert!(
             types.windows(2).all(|x| x[0] < x[1]),
-            "type info unsorted or contains duplicates"
+            "attempted to allocate an entity with duplicate components ({})",
+            core::any::type_name::<B>()
```
Ralith, line 54: To clarify, the name of *the duplicated component type* should be printed. The name of the bundle isn't very useful, and is already available in the backtrace.

src/archetype.rs
```diff
@@ -43,11 +43,21 @@ pub struct Archetype {
...
-    pub(crate) fn new(types: Vec<TypeInfo>) -> Self {
+    pub(crate) fn empty() -> Self {
+        Self::new_(Vec::new())
+    }
+
+    pub(crate) fn new<B: crate::DynamicBundle>(types: Vec<TypeInfo>) -> Self {
         debug_assert!(
             types.windows(2).all(|x| x[0] < x[1]),
-            "type info unsorted or contains duplicates"
+            "attempted to allocate an entity with duplicate components ({})",
```
Ralith, line 53: This assert is conflating two separate conditions: misordering, which is a hecs bug, and duplicate components, which are a user bug. They should be separate asserts, with separately worded errors.

macros/src/lib.rs
```diff
@@ -102,52 +102,22 @@ fn gen_bundle_impl(
...
-        quote! {
-            #[allow(non_camel_case_types)]
-            fn with_static_ids<__hecs__T>(f: impl ::std::ops::FnOnce(&[::std::any::TypeId]) -> __hecs__T) -> __hecs__T {
-                f(&#with_static_ids_inner)
-            }
-        }
-    };
-    quote! {
-        impl #impl_generics ::hecs::Bundle for #ident #ty_generics #where_clause {
-            #with_static_ids
```
Ralith: I'd (weakly) prefer to retain the separate case for monomorphic structs, so this change doesn't slow down existing code.

Ralith (changes requested):

src/archetype.rs
Ralith: Also, the equality check probably needs to be a non-debug `assert` for soundness.

src/archetype.rs
Veykril: Oh I'm sorry, I misread what you've written before, this of course makes a lot more sense.

macros/src/lib.rs
Veykril: I don't quite get what you mean. Are you saying you want to retain the lazy_static check in the non generic version? Because otherwise now the snippets for both generic and non-generic structs are the same.

Veykril:
Alright I hope I finally addressed everything. Really sorry for making you have to review more than necessary 😅

Veykril:
One other small thing I noticed, the proc-macro currently assumes `std` to be always there, should this maybe also be feature gated so that it uses `alloc::vec` instead of `std::vec` and otherwise `core` paths instead of `std`?

Ralith (approved):
Thanks, this looks great!

> Really sorry for making you have to review more than necessary sweat_smile

No worries, still easier than writing it myself :grinning:

> the proc-macro currently assumes std to be always there

Good catch; this should be fixed, but we can leave it for a separate PR.

PR #110, opened 2020-11-08
derive(Query) support

Ralith:
Monomorphic structs only, but otherwise fairly unrestricted.


Ralith:
Had to strip out the expected error for wrong_lifetime.rs due to rustc beta changing the error format and https://github.com/actions/toolkit/issues/399 not letting us gracefully ignore that. Hopefully a short term issue.

PR #111, opened 2020-11-08
Methods for querying from a mutable borrow

Ralith:
These allow dynamic borrow checks to be bypassed when there's no concurrency needed. Confusingly, despite being literally the same iterator minus some setup work, the benchmark is ~50% slower. @mjhostet, am I doing something stupid here?


Ralith:
`#[inline(always)]` to the rescue. Which is confusing, since it seemed to be getting inlined regardless, but :shrug:.

Issue #60, opened 2020-04-14
Columnar bulk insertions


Ralith:
Initialization of components from bulk column-formatted data could allow for even faster entity construction than `spawn_batch`. It's unclear if there's a real need for that as `spawn_batch` is already blazingly fast, but it could be fun.

Ralith:
On review, a `spawn_batch` with an iterator that reads from columnar data is probably going to have most of the hypothetical advantages already.  Unclear if there's really room for improvement here, especially in lieu of a compelling concrete use case.

Ralith:
The problem with `spawn_batch` is that it's static. Columnar deserialization, where the combination of components for a given batch is not known in advance, could be enabled with an interface something like (names subject to bikeshedding):

```rust
struct BatchHeader { ... }
impl BatchHeader {
    pub fn register<T: Component>(&mut self);
    pub fn allocate(self, n: usize) -> Batch;
}

struct Batch { ... }
impl Batch {
    pub fn push<T: Component>(&mut self, x: T) -> Result<(), MissingComponent>;
}
```

`Batch`'s allocations could be binary-compatible with `Archetype`, allowing practically free insertion of a fully populated `Batch` into a `World`, with a graceful slow-path or an error for incompletely populated batches. A serde deserializer could then prepare a header by mapping IDs to types, then use `DeserializeSeed` to populate a `Batch` without any extraneous copying.

PR #118, opened 2020-11-25
Serialization convenience

Ralith:
hecs does not currently provide any tools to simplify serialization. I have personally argued that serialization done right tends to be highly application-specific, as users should take care to exclude runtime-only information like GPU resource handles, incorporate non-ECS data gracefully, and elide generation information from serialized entity IDs. In my own projects, I often convert entities or even entire worlds into a more serde-friendly format and serialize that, though this is suboptimal.

Users have frequently asked what the best approach for serialization with hecs is, and I set out to address that with an example. It became clear that any easily-implemented serialization scheme would be inefficient despite significant boilerplate, and an efficient scheme, avoiding redundant copies, was complex. This convinced me that providing support to bypass the bulk of this complexity for most users is justified. While advanced users may still want to consider a custom pipeline, we should be able to dramatically lower the barrier to entry.

I originally targeted a minimal, opinionated, stateful interface along the lines of:
```rust
impl WorldSerializer {
    fn register<T: Component + Serialize, K: Serialize + 'static>(&mut self, key: K);
    fn serialize<S: serde::Seriralizer>(&self, world: &World, serializer: S);
}
```
Supporting any static `Serializer` with this pattern proved difficult, so a lower-level, more verbose approach requiring the application to implement traits was taken (see rustdocs in the diff for examples). The result is still much easier than implementing an equivalently efficient serialization pipeline from scratch, and provides better support for custom logic, particularly involving external state.

This could be expanded in the future with additional traits to allow customizing the process more deeply (e.g. to avoid baking in `Entity` serialization), though we must take care not to complicate simple applications.

As always, feedback very welcome!

### DIFF
--- a/src/serialize.rs
+++ b/src/serialize.rs
@@ -1,0 +1,399 @@
+//! Convenience tools for serializing [`World`]s; requires the `serde` feature
+//!
+//! [`Component`]s are not necessarily serializable, so we cannot directly implement [`Serialize`]
+//! for [`World`]. The helper functions defined in this module allow serialization and
+//! deserialization to rely on purpose-defined traits to control the procedures explicitly.
+//!
+//! This module builds on the public [`World::iter()`] and [`World::spawn_at()`] APIs, and are
+//! somewhat opinionated. For some applications, a more custom approach may be preferable.
+//!
+//! In terms of the serde data model, we treat a [`World`] as a map of entity IDs to user-controlled
+//! maps of component IDs to data. Backwards-incompatible changes to this model are subject to the
+//! same semantic versioning stability guarantees as the hecs API.
+
+use core::{cell::RefCell, fmt};
+
+use serde::{
+    de::{DeserializeSeed, MapAccess, Visitor},
+    ser::SerializeMap,
+    Deserializer, Serialize, Serializer,
+};
+
+use crate::{Component, EntityBuilder, EntityRef, World};
+
+/// Implements serialization of individual entities
+///
+/// Data external to the [`World`] can be exposed during serialization by storing references inside
+/// the struct implementing this trait.
+///
+/// # Example
+///
+/// ```
+/// # use serde::{Serialize, Deserialize};
+/// # #[derive(Serialize)]
+/// # struct Position([f32; 3]);
+/// # #[derive(Serialize)]
+/// # struct Velocity([f32; 3]);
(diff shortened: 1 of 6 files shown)


Ralith:
One option for further reducing boilerplate is a declarative macro. However, the boilerplate is modest enough that learning a custom DSL is not an obvious improvement.

Ralith:
Also of note is that this is row-oriented serialization. That's great for obviousness and human-friendly formats, but is not optimally efficient. An alternative would be archetype-oriented serialization, where the serialized form closely resembles the in-memory layout of a world, storing each archetype's components in columnar layout. That would improve compressibility dramatically and could improve load times, but would be much less suited to use with hand-edited data. There's a case for supporting both, so we could consider archetype-oriented serialization an area for future work.

src/serialize.rs
```diff
@@ -0,0 +1,397 @@
+//! Convenience tools for serializing [`World`]s; requires the `serde` feature
+//!
+//! [`Component`]s are not necessarily serializable, so we cannot directly implement [`Serialize`]
+//! for [`World`]. The helper functions defined in this module allow serialization and
+//! deserialization to rely on purpose-defined traits to control the procedures explicitly.
+//!
+//! These tools use public hecs APIs, and are somewhat opinionated. For some applications, a more
+//! custom approach may be preferable.
```
icefoxen, line 8: An extra sentence here pointing towards the API's for implementing that custom approach might be useful.

src/serialize.rs
```diff
@@ -0,0 +1,397 @@
...
+/// # Example
+/// ```
+/// # use serde::{Serialize, Deserialize};
+/// # #[derive(Deserialize)]
+/// # struct Position([f32; 3]);
+/// # #[derive(Deserialize)]
+/// # struct Velocity([f32; 3]);
+/// use hecs::{*, serialize::*};
+///
+/// struct Context;
```
icefoxen, line 137: `Context` here does nothing, which makes it a little harder to grok why this trait is the way it is.

Ralith:
Tweaked the docs a bit to address comments.

Ralith:
I'm debating moving this into a `hecs::serialization::row` module, to account for future introduction of columnar serialization per recent musings in #60.

Ralith:
Will pursue that in a future PR.

PR #120, opened 2020-12-05
Columnar component access

### LINKED ISSUE #60: Columnar bulk insertions

Ralith:
Fixes #60. Allows for cache-efficient column-oriented custom operations such as (de)serialization.


AngelOfSol:
+1 for the columnar API.

PR #123, opened 2020-12-16
Adds API surface to EntityBuilder to check and modify the current components of the builder.

AngelOfSol:
This is a pretty basic PR.  I contemplate adding `remove` functionality, but it was a lot more involved than this, and while I can think of use-cases I didn't find it very pressing to implement.  Let me know if you want me to update the PR with that functionality though.

### DIFF
--- a/src/entity_builder.rs
+++ b/src/entity_builder.rs
@@ -105,6 +105,36 @@ impl EntityBuilder {
         self
     }
 
+    /// Checks to see if the component of type `T` exists
+    pub fn has<T: Component>(&self) -> bool {
+        self.indices.contains_key(&TypeId::of::<T>())
+    }
+
+    /// Borrow the component of type `T`, if it exists
+    pub fn get<T: Component>(&self) -> Option<&T> {
+        let index = self.indices.get(&TypeId::of::<T>())?;
+        let (_, offset) = self.info[*index];
+        unsafe {
+            let storage = self.storage.as_ptr().add(offset).cast::<T>();
+            Some(&*storage)
+        }
+    }
+
+    /// Uniquely borrow the component of type `T`, if it exists
+    pub fn get_mut<T: Component>(&mut self) -> Option<&mut T> {
+        let index = self.indices.get(&TypeId::of::<T>())?;
+        let (_, offset) = self.info[*index];
+        unsafe {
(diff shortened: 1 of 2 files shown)


src/entity_builder.rs
```diff
@@ -104,6 +104,33 @@ impl EntityBuilder {
         }
         self
     }
+    /// Checks to see if the component of type `T` exists
```
Ralith, line 107: Empty line between function definitions, please.

src/entity_builder.rs
```diff
@@ -104,6 +104,33 @@ impl EntityBuilder {
...
+    pub fn get<T: Component>(&self) -> Option<&T> {
+        let index = self.indices.get(&TypeId::of::<T>())?;
+        let (_, offset) = self.info[*index];
+        unsafe {
+            let storage = self.storage.as_ptr().add(offset).cast::<T>();
+            Some(&*storage)
+        }
+    }
+    /// Uniquely borrow the component of type `T`, if it exists
+    pub fn get_mut<T: Component>(&self) -> Option<&mut T> {
```
Ralith, line 121:
```suggestion
    pub fn get_mut<T: Component>(&mut self) -> Option<&mut T> {
```

tests/tests.rs
```diff
@@ -142,6 +142,20 @@ fn build_entity() {
...
+    assert!(entity.has::<i32>());
+    assert!(!entity.has::<usize>());
+    assert_eq!(*entity.get::<&str>().unwrap(), "ghi");
+    assert_eq!(*entity.get::<i32>().unwrap(), 123);
+    assert_eq!(entity.get::<usize>(), None);
+    *entity.get_mut::<i32>().unwrap() = 456;
+    assert_eq!(*entity.get::<i32>().unwrap(), 456);
+    let g = world.spawn(entity.build());
+    assert_eq!(*world.get::<&str>(g).unwrap(), "ghi");
+    assert_eq!(*world.get::<i32>(g).unwrap(), 456);
```
Ralith, line 158: This should be a separate test.

Ralith (changes requested):
Thanks! I don't mind leaving remove for future work, assuming anyone actually wants it.

AngelOfSol:
Updated with respect to your comments, let me know if theres anything else.

PR #124, opened 2020-12-16
Implements out of band cloning for World

AngelOfSol:
User needs to keep a registry of functions to use to clone the world, and the cloning can and will
fail if theres an outstanding unique borrow, or the registry is missing
a component type.

Outstanding questions:

How should I handle the two error cases?  I wanted to return a nice error, but if you turn off debug_asserts, then the current method doesn't work.  I can switch both to panicking, I'm guessing I should print out a more accurate panic when the info is available though. [so this can't be merged as is]

If keeping the Result is intended, is there anything I should do on the borrow/release section?  If clone_with fails to borrow here, it won't release the borrows its acquired so far.

Also, someone needs to take a looksie to make sure I'm not breaking invariants with my unsafe code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ maintenance = { status = "actively-developed" }
 [features]
 default = ["std"]
 std = []
+clone = []
 # Enables derive(Bundle)
 macros = ["hecs-macros", "lazy_static"]
 
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -8,7 +8,9 @@
 use bencher::{benchmark_group, benchmark_main, Bencher};
 use hecs::*;
 
+#[derive(Clone)]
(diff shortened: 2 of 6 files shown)


AngelOfSol:
Adjusted the API a bit, so that the CloneRegistry is passed in on World creation and kept around.

AngelOfSol:
Did some more fixes to the unsafe portion of the code, after testing it in an actual usage scenario (my game).  Found a bunch of buggy stuff so I the fixes I reworked should have fixed those issues.

Remaining concerns are naming, whether or not I'm handling the borrow/release properly, and if Entities::clone is doing its atomics properly.

AngelOfSol:
Closing because the future columnar API supports this usecase faster and clearer than the manually written unsafe code.

PR #134, opened 2021-02-09
Fix length tracking for unallocated reserved entities

### LINKED ISSUE #133: reserve_entity does not increment entities.len() when flushed
Repro:

```rust
#[test]
fn test() {
    let mut e = Entities::default();
    let e1 = e.reserve_entity();
    e.flush(|x, y| {});
    // this fails
    assert_eq!(e.len(), 1);
}
```

This causes "subtract with overflow" errors when despawning a reserved entity (after flushing).

```rust
#[test]
fn reserve_test() {
    let mut world = World::new();
    let entity = world.reserve_entity();
    world.flush();
    world.despawn(entity).unwrap();
}
```

This effectively makes entity reservation useless (unless the entities are never despawed). I'm experimenting with using the new hecs entity allocator in bevy, which is what surfaced this.

Introduced by #86

@Ralith  @mjhostet

Ralith:
Fixes #133. Looks like this was a pretty simple omission.

@mjhostet would love a sanity check from you!

### DIFF
--- a/macros/src/bundle.rs
+++ b/macros/src/bundle.rs
@@ -185,7 +185,7 @@ fn struct_fields(fields: &syn::Fields) -> (Vec<&syn::Type>, Vec<syn::Member>) {
     }
 }
 
-fn member_as_idents<'a>(members: &'a [syn::Member]) -> Vec<Cow<'a, syn::Ident>> {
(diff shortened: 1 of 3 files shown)


mjhostet:
@Ralith Looks good, thanks!

PR #138, opened 2021-02-15
Adds an example of using the new columnar API to clone world.

AngelOfSol:
Let me know if you think I should make any changes to the example.  I think it's a pretty good reference example of what you can do with the new API, though I'm curious if some sort of MapWorld<> API is reasonable.

### DIFF
--- a/examples/columnar_clone.rs
+++ b/examples/columnar_clone.rs
@@ -1,0 +1,142 @@
+use core::any::TypeId;
+use std::ops::{Deref, DerefMut};
+
+use hecs::{Archetype, ColumnBatchBuilder, ColumnBatchType, Component, World};
+
+fn main() {
+    let mut world = CloneableWorld {
+        inner: World::new(),
+        clone_registry: CloneRegistry::default()
+            .register::<String>()
+            .register::<u32>(),
+    };
+
+    world.spawn((4u32,));
+    world.spawn((8u32, "test".to_string()));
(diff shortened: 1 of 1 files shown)


examples/columnar_clone.rs
```diff
@@ -0,0 +1,142 @@
...
+    CloneEntry {
+        type_id: TypeId::of::<T>(),
+        add_type: |batch_type| {
+            batch_type.add::<T>();
+        },
+        add_values: |batch, arch| {
+            let mut writer = batch.writer::<T>().unwrap();
+            for item in arch.get::<T>().unwrap().iter() {
+                if writer.push(item.clone()).is_err() {
+                    panic!()
```
Ralith, line 77:
```suggestion
                    unreachable!()
```

examples/columnar_clone.rs
```diff
@@ -0,0 +1,142 @@
+use core::any::TypeId;
```
Ralith, line 1: We should include module-level doc comments that explain what this demonstrates, why it's useful, why special effort is needed (there's no static guarantee that all components are `Clone`) and what the caveats are (unregistered components aren't cloned).

Ralith (changes requested):

examples/columnar_clone.rs
AngelOfSol: Cool, I'll get on (I've not written an example before)

Issue #154, opened 2021-04-01
Prepared queries


Ralith:
Query start-up time could be reduced by cacheing the list of archetypes to traverse, either in a `HashMap` analogous to how spawns/inserts/removes are cached, or in a structure that the user keeps track of themselves.

PR #157, opened 2021-05-02
Use slice iterator in QueryIter implementation

adamreichold:
I am currently trying to figure out how to implement #154 and noticed that `QueryIter::len` seems to yield the overall number of items instead of yielding only those items remaining as required by the contract of [Iterator::size_hint](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.size_hint). I tried fixing this by changing the implementation to use a slice iterator (which also reduces the size of `QueryIter`) and considering the remaining items of the current chunk.

I then made the same modification for `BatchedIter` as a separate commit to keep things consistent, even though I do not think it necessarily improves that implementation.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -7,6 +7,7 @@
 
 use core::marker::PhantomData;
 use core::ptr::NonNull;
+use core::slice::Iter as SliceIter;
 
 use crate::archetype::Archetype;
 use crate::entities::EntityMeta;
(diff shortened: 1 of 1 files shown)


adamreichold:
@googlebot I signed it!

adamreichold:
I do not yet understand why the two variants behave differently, but this changes does seem to improve querying via shared borrows slight on my machine, i.e. from
```
test iterate_100k     ... bench:      54,560 ns/iter (+/- 440)
test iterate_mut_100k ... bench:      54,580 ns/iter (+/- 536)
```
on `master` to
```
test iterate_100k     ... bench:      41,029 ns/iter (+/- 246)
test iterate_mut_100k ... bench:      54,439 ns/iter (+/- 374)
```
with this change.

Ralith:
Nice catch, thanks!

> this changes does seem to improve querying via shared borrows slight on my machine

Weird, but I'll take it!

Ralith (approved):

PR #158, opened 2021-05-02
Consistently apply TypeIdMap optimization where possible

adamreichold:
While looking into #154, I noticed that `TypeIdMap` was used in some but not all possible places where hash tables keyed by type ID were used.

### DIFF
--- a/src/archetype.rs
+++ b/src/archetype.rs
@@ -36,7 +36,7 @@ pub struct Archetype {
     data_size: usize,
     /// Maps static bundle types to the archetype that an entity from this archetype is moved to
     /// after removing the components from that bundle.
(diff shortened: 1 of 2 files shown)


adamreichold:
I guess the following numbers could be interpreted as a slight improve for insertion/removal, e.g. from
```
test insert           ... bench:      87,791 ns/iter (+/- 1,016)
test remove           ... bench:     102,749 ns/iter (+/- 1,652)
```
to
```
test insert           ... bench:      84,697 ns/iter (+/- 789)
test remove           ... bench:      96,803 ns/iter (+/- 1,305)
```
but the errors are pretty large.

Ralith (approved):
Good catch, thanks!

Issue #69, opened 2020-08-02
Dynamically typed queries


Ralith:
hecs could be extended with support for executing queries constructed at runtime rather than specified statically in types. This could be useful for e.g. interactively user-specified queries or for running queries in embedded scripting environments.

An implementation could look something like:

```rust
impl World {
    fn query_dynamic(&self, q: DynamicQuery) -> DynamicQueryBorrow<'_>;
}

pub enum DynamicQuery {
    Get(TypeId),
    GetMut(TypeId),
    And(Vec<DynamicQuery>),
}
```

where `DynamicQueryBorrow<'_>` contains an iterator which yields `DynamicQueryItem` which has

```
impl DynamicQueryItem<'_> {
    fn get<T: Component>(&self) -> Option<&T>;
    fn get_mut<T: Component>(&self) -> Option<&mut T>;
}
```

This is complex enough that I'm not likely to work on it unless there's significant demand, so please leave a comment if you're interested!

Grinshpon:
This is definitely interesting. I don't have an ultra-specific use case, but I'm drawing out a framework I'd like to make that utilizes the advantages of an ECS architecture but still provides some form of scripting component for fast iteration and smoother workflow.

Plus it'd be nice to get around Rust's compile times by writing non-performance-critical systems in a script.

Ralith:
Concrete use cases are particularly important because they can help inform the exact shape the API could take to solve a specific problem.

Grinshpon:
I realize that. When I can write out a concrete case I'll come back to this issue to describe it.

xacrimon:
I'm quite interested in this. I'd like to experiment with hooking hecs into the scripting language in my engine.

Ralith:
Bevy is exploring [a similar effort](https://github.com/bevyengine/bevy/pull/623), which may at least provide inspiration. Of particular interest is the introduction of dynamic *components*, treated as plain fixed-length data associated with a custom ID.

A key question for proceeding with API design here is what's required for passing results to a scripting layer easily and efficiently. Examples of the patterns needed to provide low-overhead and ergonomic access to Rust data from a script would be valuable.

Ralith:
Another compelling idea from bevy: define a space of `ComponentId`s independent of `TypeId`s, map from `TypeId`s to `ComponentId`s for conventional use, and allow explicit allocation of `ComponentId`s for dynamic cases.

sdleffler:
[...]

PR #159, opened 2021-05-02
Introduce an out-of-line query cache storing matching archetypes

### LINKED ISSUE #154: Prepared queries

adamreichold:
This uses a separately managed query cache that checks both the memory address of the archetypes as well as the generation number to ensure that the cache its keeping belongs to the same world and in the same state as the query it is backing. (There might be an ABA problem here, e.g. world A adds archetypes until generation N, caches are query, adds more archetypes and reallocates that `Vec`, then world B adds archetypes until generation N which are backed by the same allocation that was dropped by world A after caching a query. Uniquely tagging worlds using an `AtomicUsize` instead might resolve this.)

The main issue I currently see is that the performance does not seem to improve measurably, e.g.
```
test iterate_cached_100k_by_50   ... bench:      22,605 ns/iter (+/- 556)
test iterate_uncached_100k_by_50 ... bench:      23,051 ns/iter (+/- 490)
```
so either I am doing something wrong in the implementation or the complexity implied by the caching is not worth the gain.

Fixes #154

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -107,78 +107,57 @@ fn iterate_mut_100k(b: &mut Bencher) {
     })
 }
 
-fn spawn_100k_by_50(world: &mut World) {
-    for i in 0..2_000 {
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 0]));
-        world.spawn((Position(-(i as f32)), [(); 0]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 1]));
-        world.spawn((Position(-(i as f32)), [(); 1]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 2]));
-        world.spawn((Position(-(i as f32)), [(); 2]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 3]));
-        world.spawn((Position(-(i as f32)), [(); 3]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 4]));
-        world.spawn((Position(-(i as f32)), [(); 4]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 5]));
-        world.spawn((Position(-(i as f32)), [(); 5]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 6]));
-        world.spawn((Position(-(i as f32)), [(); 6]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 7]));
-        world.spawn((Position(-(i as f32)), [(); 7]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 8]));
-        world.spawn((Position(-(i as f32)), [(); 8]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 9]));
-        world.spawn((Position(-(i as f32)), [(); 9]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 10]));
-        world.spawn((Position(-(i as f32)), [(); 10]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 11]));
-        world.spawn((Position(-(i as f32)), [(); 11]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 12]));
-        world.spawn((Position(-(i as f32)), [(); 12]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 13]));
-        world.spawn((Position(-(i as f32)), [(); 13]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 14]));
-        world.spawn((Position(-(i as f32)), [(); 14]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 15]));
-        world.spawn((Position(-(i as f32)), [(); 15]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 16]));
-        world.spawn((Position(-(i as f32)), [(); 16]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 17]));
-        world.spawn((Position(-(i as f32)), [(); 17]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 18]));
-        world.spawn((Position(-(i as f32)), [(); 18]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 19]));
-        world.spawn((Position(-(i as f32)), [(); 19]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 20]));
-        world.spawn((Position(-(i as f32)), [(); 20]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 21]));
-        world.spawn((Position(-(i as f32)), [(); 21]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 22]));
-        world.spawn((Position(-(i as f32)), [(); 22]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 23]));
-        world.spawn((Position(-(i as f32)), [(); 23]));
-        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 24]));
-        world.spawn((Position(-(i as f32)), [(); 24]));
+fn spawn_100_by_50(world: &mut World) {
+    fn spawn_two<const N: usize>(world: &mut World, i: i32) {
+        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); N]));
+        world.spawn((Position(-(i as f32)), [(); N]));
+    }
+
+    for i in 0..2 {
+        spawn_two::<0>(world, i);
+        spawn_two::<1>(world, i);
+        spawn_two::<2>(world, i);
+        spawn_two::<3>(world, i);
+        spawn_two::<4>(world, i);
+        spawn_two::<5>(world, i);
+        spawn_two::<6>(world, i);
+        spawn_two::<7>(world, i);
+        spawn_two::<8>(world, i);
+        spawn_two::<9>(world, i);
+        spawn_two::<10>(world, i);
+        spawn_two::<11>(world, i);
+        spawn_two::<12>(world, i);
+        spawn_two::<13>(world, i);
+        spawn_two::<14>(world, i);
+        spawn_two::<15>(world, i);
+        spawn_two::<16>(world, i);
+        spawn_two::<17>(world, i);
+        spawn_two::<18>(world, i);
+        spawn_two::<19>(world, i);
+        spawn_two::<20>(world, i);
+        spawn_two::<21>(world, i);
+        spawn_two::<22>(world, i);
+        spawn_two::<23>(world, i);
+        spawn_two::<24>(world, i);
     }
 }
 
-fn iterate_uncached_100k_by_50(b: &mut Bencher) {
+fn iterate_uncached_100_by_50(b: &mut Bencher) {
     let mut world = World::new();
-    spawn_100k_by_50(&mut world);
+    spawn_100_by_50(&mut world);
     b.iter(|| {
-        for (_, (pos, vel)) in &mut world.query::<(&mut Position, &Velocity)>() {
+        for (_, (pos, vel)) in world.query_mut::<(&mut Position, &Velocity)>() {
             pos.0 += vel.0;
         }
     })
 }
 
-fn iterate_cached_100k_by_50(b: &mut Bencher) {
+fn iterate_cached_100_by_50(b: &mut Bencher) {
     let mut world = World::new();
-    let mut cache = world.query_cache();
-    spawn_100k_by_50(&mut world);
+    spawn_100_by_50(&mut world);
+    let mut query = world.query::<(&mut Position, &Velocity)>().prepare(&world);
     b.iter(|| {
-        let mut query = world.query::<(&mut Position, &Velocity)>();
-        for (_, (pos, vel)) in query.iter_cached(&mut cache) {
+        for (_, (pos, vel)) in query.iter(&mut world) {
             pos.0 += vel.0;
         }
     })
@@ -202,8 +181,8 @@ benchmark_group!(
     insert,
     iterate_100k,
     iterate_mut_100k,
-    iterate_uncached_100k_by_50,
-    iterate_cached_100k_by_50,
+    iterate_uncached_100_by_50,
+    iterate_cached_100_by_50,
     build
 );
 benchmark_main!(benches);
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -156,8 +156,31 @@ fn iterate_cached_100_by_50(b: &mut Bencher) {
     let mut world = World::new();
     spawn_100_by_50(&mut world);
     let mut query = world.query::<(&mut Position, &Velocity)>().prepare(&world);
+    let _ = query.borrow(&world).iter();
     b.iter(|| {
-        for (_, (pos, vel)) in query.borrow(&mut world).iter() {
+        for (_, (pos, vel)) in query.borrow(&world).iter() {
+            pos.0 += vel.0;
+        }
+    })
+}
+
+fn iterate_mut_uncached_100_by_50(b: &mut Bencher) {
+    let mut world = World::new();
+    spawn_100_by_50(&mut world);
+    b.iter(|| {
+        for (_, (pos, vel)) in world.query_mut::<(&mut Position, &Velocity)>() {
+            pos.0 += vel.0;
+        }
+    })
+}
+
+fn iterate_mut_cached_100_by_50(b: &mut Bencher) {
+    let mut world = World::new();
+    spawn_100_by_50(&mut world);
+    let mut query = world.query::<(&mut Position, &Velocity)>().prepare(&world);
+    let _ = query.iter_mut(&mut world);
+    b.iter(|| {
+        for (_, (pos, vel)) in query.iter_mut(&mut world) {
             pos.0 += vel.0;
         }
     })
@@ -183,6 +206,8 @@ benchmark_group!(
     iterate_mut_100k,
     iterate_uncached_100_by_50,
     iterate_cached_100_by_50,
+    iterate_mut_uncached_100_by_50,
+    iterate_mut_cached_100_by_50,
     build
 );
 benchmark_main!(benches);
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -156,9 +156,9 @@ fn iterate_cached_100_by_50(b: &mut Bencher) {
     let mut world = World::new();
     spawn_100_by_50(&mut world);
     let mut query = world.query::<(&mut Position, &Velocity)>().prepare(&world);
-    let _ = query.borrow(&world).iter();
+    let _ = query.query(&world).iter();
     b.iter(|| {
-        for (_, (pos, vel)) in query.borrow(&world).iter() {
+        for (_, (pos, vel)) in query.query(&world).iter() {
             pos.0 += vel.0;
         }
     })
@@ -178,9 +178,9 @@ fn iterate_mut_cached_100_by_50(b: &mut Bencher) {
     let mut world = World::new();
     spawn_100_by_50(&mut world);
     let mut query = world.query::<(&mut Position, &Velocity)>().prepare(&world);
-    let _ = query.iter_mut(&mut world);
+    let _ = query.query_mut(&mut world);
     b.iter(|| {
-        for (_, (pos, vel)) in query.iter_mut(&mut world) {
+        for (_, (pos, vel)) in query.query_mut(&mut world) {
             pos.0 += vel.0;
         }
     })
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -155,7 +155,7 @@ fn iterate_uncached_100_by_50(b: &mut Bencher) {
 fn iterate_cached_100_by_50(b: &mut Bencher) {
     let mut world = World::new();
     spawn_100_by_50(&mut world);
-    let mut query = world.query::<(&mut Position, &Velocity)>().prepare(&world);
+    let mut query = PreparedQuery::<(&mut Position, &Velocity)>::default();
     let _ = query.query(&world).iter();
     b.iter(|| {
         for (_, (pos, vel)) in query.query(&world).iter() {
@@ -177,7 +177,7 @@ fn iterate_mut_uncached_100_by_50(b: &mut Bencher) {
 fn iterate_mut_cached_100_by_50(b: &mut Bencher) {
     let mut world = World::new();
     spawn_100_by_50(&mut world);
-    let mut query = world.query::<(&mut Position, &Velocity)>().prepare(&world);
+    let mut query = PreparedQuery::<(&mut Position, &Velocity)>::default();
     let _ = query.query_mut(&mut world);
     b.iter(|| {
         for (_, (pos, vel)) in query.query_mut(&mut world) {
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -146,7 +146,7 @@ fn iterate_uncached_100_by_50(b: &mut Bencher) {
     let mut world = World::new();
     spawn_100_by_50(&mut world);
     b.iter(|| {
-        for (_, (pos, vel)) in world.query_mut::<(&mut Position, &Velocity)>() {
+        for (_, (pos, vel)) in world.query::<(&mut Position, &Velocity)>().iter() {
             pos.0 += vel.0;
         }
     })
@@ -157,7 +157,7 @@ fn iterate_cached_100_by_50(b: &mut Bencher) {
     spawn_100_by_50(&mut world);
     let mut query = world.query::<(&mut Position, &Velocity)>().prepare(&world);
     b.iter(|| {
-        for (_, (pos, vel)) in query.iter(&mut world) {
+        for (_, (pos, vel)) in query.borrow(&mut world).iter() {
             pos.0 += vel.0;
         }
     })
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -107,6 +107,83 @@ fn iterate_mut_100k(b: &mut Bencher) {
     })
 }
 
+fn spawn_100k_by_50(world: &mut World) {
+    for i in 0..2_000 {
+        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 0]));
+        world.spawn((Position(-(i as f32)), [(); 0]));
+        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 1]));
+        world.spawn((Position(-(i as f32)), [(); 1]));
+        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 2]));
+        world.spawn((Position(-(i as f32)), [(); 2]));
+        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 3]));
+        world.spawn((Position(-(i as f32)), [(); 3]));
+        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 4]));
+        world.spawn((Position(-(i as f32)), [(); 4]));
+        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 5]));
+        world.spawn((Position(-(i as f32)), [(); 5]));
+        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 6]));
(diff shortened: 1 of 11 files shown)


Ralith:
Thanks for having a look at this! Interesting results. On my laptop, it's even worse--the cached is ~34μs, vs. ~27μs. However, if I drop the number of entities per archetype to 2, things flip around: 1.9μs cached, 2.2μs uncached. So, this looks like a higher base cost, more efficient per-archetype cost situation.

"Large number of small archetypes" isn't a totally unrealistic case, so this might be worthwhile if we can get the base cost down. I haven't looked at profiling, but my first idea is to replace `QueryCache` structure with a `PreparedQuery<Q>` structure that's obtained from a world for a specific query. That saves a hash table lookup, at least, and it's hard to imagine how that would be slower.

adamreichold:
> I haven't looked at profiling

I did have a loot at it using Callgrind but I did not find anything useful at the function level and I have not yet looked at the instruction level.

> my first idea is to replace QueryCache structure with a PreparedQuery<Q> structure that's obtained from a world for a specific query. That saves a hash table lookup, at least, and it's hard to imagine how that would be slower.

I will give it a try, but it will probably take me a while to find some spare time...

adamreichold:
> my first idea is to replace QueryCache structure with a PreparedQuery" structure that's obtained from a world for a specific query. That saves a hash table lookup, at least, and it's hard to imagine how that would be slower."

I did implement a rough version of this but it seems like it does not seem to help much, i.e. with 2 entities per 50 archetypes:
```
test iterate_cached_100_by_50   ... bench:         328 ns/iter (+/- 30)
test iterate_uncached_100_by_50 ... bench:         551 ns/iter (+/- 12)
```
and for 2000 entities per 50 archetypes, the uncached version was faster for me as well.

Maybe as a data point of why it isn't faster, I found the following result
```
test iterate_cached_100_by_50   ... bench:          44 ns/iter (+/- 1)
test iterate_uncached_100_by_50 ... bench:         560 ns/iter (+/- 19)
```
when I not just cache the relevant archetypes, but that and basically the result of `Q::Fetch::new(archetype).unwrap()`, i.e the data pointers themselves.

But AFAIU, this would mean that the prepared query is invalidated whenever entities are spawned or at least whenever any archetype is reallocated which does seem to be tracked by the archetypes generation. Furthermore, this might not be useful for reasonable use cases as a stable set of archetypes seems plausible whereas a stable set of entities appears to be a rather special case. @Ralith Do you think it would be worthwhile to track reallocation of archetypes and invalidate prepared queries when that happens?

Ralith:
> when I not just cache the relevant archetypes, but that and basically the result of Q::Fetch::new(archetype).unwrap(), i.e the data pointers themselves.

Ooh, good idea! That saves the per-archetype hash table lookup, too.

> But AFAIU, this would mean that the prepared query is invalidated whenever entities are spawned or at least whenever any archetype is reallocated

Right, the pointers will be invalidated on every realloc, which is a bit annoying. I think we could freshness by cacheing the base `data` pointer of the archetype and validating it. I do actually think this would work okay in practice, because spawning/despawning only reallocs if more storage is required, and we use `Vec`-style exponential realloc to ensure that any sane system reaches a steady state on a reasonable amount of time, at which point no further invalidation occurs.

That said, I think we could avoid invalidation on realloc entirely with some modest internal changes. Rather than storing the actual pointer to component data, we could store information sufficient to efficiently determine the offset of component data within the allocation, based on some key properties:
- The order of components within a single archetype is fixed
- The amount of space allocated for a component `T` in an archetype is always `size_of::<T>() * self.capacity()`
- Because components are ordered by descending alignment, there is no padding between them

Given these, we can infer that a component's offset in an archetype's allocation is always exactly `self.capacity()` multiplied by the sum of the sizes of the components that precede it, so that sum should be all we need to store; the rest can be computed with trivial arithmetic and no extra indirection. Does that make sense?

I also expect the benefit of preparing queries to grow if/when we add dynamic components as proposed in https://github.com/Ralith/hecs/issues/69#issuecomment-770095884, which would add an additional layer of hashing for typical use.

adamreichold:
> Given these, we can infer that a component's offset in an archetype's allocation is always exactly self.capacity() multiplied by the sum of the sizes of the components that precede it, so that sum should be all we need to store; the rest can be computed with trivial arithmetic and no extra indirection. Does that make sense?

I have though about this and did some experimentation and I have to admit that I cannot follow it. If I store the sum per archetype and component within it, why not just store the base pointer itself? And isn't that sum (since it depends on the capacity) also invalidated when the archetype is reallocated?

But more generally, I think there is a simpler way by combining this PR and #160: If the offsets/borrows are stored in a (sorted) list, we can cache the indices into that list in the prepared query. They will stay stable as along as the archetype generation is unchanged, so the invalidation is exactly as it is already implemented. For actually executing the query, we just take these indices to retrieve the offsets/borrows (which must access again in any case) without a "look-up", i.e. binary search or hash table look-up. The main change required to do this is to adjust `Archetype::{get_base,borrow,borrow_mut,release}` to accept numeric indices and `Fetch::access` to produce them so they can be cached for future use.

adamreichold:
> I think there is a simpler way by combining this PR and #160: ...

Now that I have written it down, I probably do not need #160 after all. I could also cache raw pointers of type `*const TypeState` into the `Archetype::state` hash table and pass those around instead of indices.

Ralith:
> If I store the sum per archetype and component within it, why not just store the base pointer itself

Because the set of component types in an archetype never changes, and neither does their ordering, but the base pointer changes on realloc.

> (since it depends on the capacity)

Sorry, I meant the sum of the sizes of the component types preceding a given component in the ordered list associated with the archetype (realized in `Archetype::types`). For example, if you have an archetype for `u64`, `u32`, and `u16`, then their respective sums would be 0, 8, and 12. This is the only data that you hold onto. To compute the offset of a given component within an archetype on-demand, you then take the associated sum and multiply it by the archetype's *current* capacity.

> I could also cache raw pointers of type *const TypeState into the Archetype::state hash table and pass those around instead of indices.

That would work, since the hash table is immutable once constructed. However, I'd prefer the approach I proposed since it's safer.

Ralith:
I suppose the drawback to the size-sum-based approach is we need to do additional refactoring to expose the `AtomicBorrow` without hitting the hash table. I think it'd be reasonable to just move that into a vector with the same ordering as `types` and store indexes, which shouldn't be too bad.

adamreichold:
> I suppose the drawback to the size-sum-based approach is we need to do additional refactoring to expose the AtomicBorrow without hitting the hash table. I think it'd be reasonable to just move that into a vector with the same ordering as types and store indexes, which shouldn't be too bad.

Since I have to reference into the archetype for the borrows in any case, isn't storing pointers (be they raw pointers or indices) into the `Archetype::state` doing the above and avoiding reifing all the data layout assumptions, so overall the simpler thing to do?

I do feel icky about using raw pointers into the hash table as well since the data structure certainly feels more opaque than a resizeable array, but we need to do so for borrowing without searching in any case. So getting at the stored offsets comes for free along the same path.

Ralith:
Ah, wait, the current impl *does* actually reallocate the hash table on grow, just because it was convenient to do so. So while I sympathize that the raw pointer trick would be easier, it won't work as it stands, and the refactoring to make it work might not end up saving much effort vs. refactoring so we don't need it at all.

adamreichold:
> To be clear, there's no need to change any component data layout assumptions, we just need to move the TypeState into a linear table so borrow state can be found by index.

This is basically what I meant with "combining with #160", but my main observation is still that if I access `TypeState` without searching (now by index) than I immediately get `borrow` and `offset`, so there is no need for any other caching/computation of the data pointer.

So in summary, my proposal for this MR would be:
* turn in `Archetype::state` into an array and store the indices into this array within `Archetype::types` for looking by type ID
* turn `Fetch::new` into a two-phase process, first looking up types and returning the indices into `Archetype::state`
* a prepared query will cache those archetypes and their associated type state indices
* a transient query will just pass them directly to the second phase of `Fetch::new` (and `Fetch::borrow` which will take indices as well)
* extend `Archetype::{get_base,borrow,borrow_mut,release}` to accept type state indices
* a prepared query will check the world ID and archetypes generation of the world it asked to iterate
* if it matches, the above archetypes and type state indices will be used as-is to borrow and fetch offsets so no hash table look-up is necessary, otherwise they will be recomputed

benches/bench.rs
```diff
@@ -107,6 +107,82 @@ fn iterate_mut_100k(b: &mut Bencher) {
     })
 }
 
+fn spawn_100_by_50(world: &mut World) {
+    for i in 0..2 {
+        world.spawn((Position(-(i as f32)), Velocity(i as f32), [(); 0]));
```
adamreichold, line 112: Depending on the MSRV, this could be significantly simplified using const generics.

src/world.rs
```diff
@@ -47,15 +48,19 @@ pub struct World {
...
 impl World {
     /// Create an empty world
     pub fn new() -> Self {
+        static NEXT_ID: AtomicUsize = AtomicUsize::new(0);
+
         Self {
             entities: Entities::default(),
             archetypes: ArchetypeSet::new(),
             bundle_to_archetype: HashMap::default(),
+            id: NEXT_ID.fetch_add(1, Ordering::Relaxed),
```
adamreichold, line 63: This should probably guard against overflow like
```rust
NEXT_ID.fetch_update(Ordering::Relaxed, Ordering::Relaxed, |next_id| next_id.checked_add(1)).unwrap()
```

benches/bench.rs
Ralith: Great idea! I don't think MSRV is a concern for benchmark code at all, so go nuts.

Ralith:
> if I access TypeState without searching (now by index) than I immediately get borrow and offset, so there is no need for any other caching/computation of the data pointer.

Ah, right, good catch. I suppose we could remove `offset` entirely in favor of the sum trick, but I don't have any particular reason to believe it'd be useful to do so.

> So in summary, my proposal for this MR would be:

Overall this sounds good. A couple notes:

> turn in Archetype::state into an array and store the indices into this array within Archetype::types for look up by type ID

There shouldn't be any need for indices stored in `types`, as the arrays should have exactly the same order at all times. We would need to introduce a `TypeIdMap<usize>` to preserve the existing hash lookup behavior, though, unless benchmarking indicates that it's more trouble than it's worth to do so.

> a prepared query will check the world ID

We could avoid the need for a world ID if we `assert` that the component types a prepared query finds in its cached locations match its expectations. I'm not sure if the extra storage and comparisons would impact performance, though. I'm okay either way.

src/world.rs
Ralith: If we're relying on it for soundness, yeah. Alternatively, if we use an `AtomicU64` then it's probably reasonable to assume that overflow won't happen.

adamreichold:
> We could avoid the need for a world ID if we assert that the component types a prepared query finds in its cached locations match its expectations.

I think this would not prevent would not prevent a prepared query operating on a world which has the same archetypes in the same places but additional ones that should also match the query. Even if the performance is not an issue.

Ralith:
That's correct. I think it'd be fine to document that the results may be incomplete (but safe) if a prepared query is used on a world other than the one for which it's prepared. Switching back and forth between worlds would defeat the point of using a prepared query, after all.

adamreichold:
>  I think it'd be fine to document that the results may be incomplete (but safe) if a prepared query is used on a world other than the one for which it's prepared.

But have the caveat if it relatively simple and faster to check the unique world ID which avoids the issue entirely?

Ralith:
Yeah, world IDs are definitely simpler. Relying on global state feels a bit cheesy to me, but it's an atomic that's only hit on a slow path, so w/e. It might lead to surprising behavior if e.g. two separate cdylibs use hecs in the same process and try to share prepared queries, but that feels pretty contrived.

adamreichold:
Ok, please bear with me, I know this is still sketchy but I have pushed an implementation that does almost what we discussed but is unfinished in at least two places and does not use indices yet. But the performance seems to look like what we are aiming for:
```
test iterate_cached_100_by_50   ... bench:          57 ns/iter (+/- 4)
test iterate_uncached_100_by_50 ... bench:         534 ns/iter (+/- 19)
```

Just to make things work, I just fixed things up so `Archetype::state` is not reallocated during `Archetype::grow` and stored raw pointers into that hash table. Since `Archetype::state` should now not ever be reallocated, this should be as safe as say using indices with unchecked access.

It is unfinished because I restrict myself to the mutable prepared query case, i.e borrowing via the cached state access is not yet implemented. But I also compare this to the mutable uncached case above, so at least the numbers should be a fair comparison. I also did not fix up the derive macro for `Query` because I did not yet fully grasp how that works exactly. (This is why the tests including miri fail.) (The Clippy lint is actually gone on nightly but I will fix that when I complete the missing documentation after we settled on the approach.)

Finally, I think I should probably replace the Archetype code paths for `get_base`/`borrow`/etc. instead of adding new methods?

Ralith:
Looks like the concept's validated, nice! One note--in the spirit of safety, I'd also prefer to reference archetypes by index rather than pointer, even though we're tracking invalidation accurately through generation.

I'm all on board with minimizing the number of similar methods on `Archetype` so long as doing so doesn't lead to significant increases in complexity elsewhere, or decrease safety gratuitously. In the spirit of avoiding duplication, I wonder if we could even reuse the same iterator between prepared and immediate queries, rendering `Fetch::new` entirely obsolete, without sacrificing too much.

adamreichold:
> I'd also prefer to reference archetypes by index rather than pointer, even though we're tracking invalidation accurately through generation.

Done. Performance is unchanged, but rustc forced me to fix the inconsistent lifetimes of `PreparedQueryIter`, so certainly a good call.

adamreichold:
> In the spirit of avoiding duplication, I wonder if we could even reuse the same iterator between prepared and immediate queries, rendering Fetch::new entirely obsolete, without sacrificing too much.

I added a commit that does that but I think the price is pretty high: I had to make `QueryIter` generic of the internal type of state access which thereby becomes part of the API if we do not want reduce this to return `impl Iterator` which seems bad for downstream consumers who might want to store those iterators themselves. But that seemed to be the only to avoid forcing immediate queries to allocate a temporary buffer to store the states in. Please have a look if you think I should continue to pursue this approach.

If we just want to get rid of `Fetch::new`, it seems like it would be easier to just make its current users call both `Fetch::prepare` and `Fetch::execute` in sequence. Or we could provide a default implementation like
```rust
fn new(archetype: &'a Archetype) -> Option<Self> {
    Self::prepare(archetype).map(|state| unsafe { Self::execute(archetype, state) })
```
to make this automatic.

adamreichold:
Updated to do list:
- [ ] Implement borrow checking for prepared queries, e.g. add a `fn borrow<'q>(&'q mut self, world: &'q World) -> PreparedQueryBorrow<'q, Q>` method
- [ ] Replace access to type state via raw pointers by using indices and therefore changing the type of `Archetype::state`
- [ ] Update derive macro after finalising changes to `Fetch` trait 
- [ ] Add missing documentation

Ralith:
> But that seemed to be the only to avoid forcing immediate queries to allocate a temporary buffer to store the states in.

That's what I had in mind, actually. Some other ideas, if the allocation is a problem:
- Replace `state` with a `&mut dyn Iterator<Item=etc>`
- Replace `state` with an enum
- Define a separate iterator type, but share most of the implementation via helper functions/common member structs

I think any approach to reusing the iterator type which is more complex than just writing another one defeats the point. I also don't mind leaving this for future work.

> Updated to do list:

:+1:

adamreichold:
> I also don't mind leaving this for future work.

Honestly speaking, I feel like the series and the discussion are pretty voluminous already and hence I would prefer that option ATM. I will stash the commit and aim to propose this as a follow-up PR.

adamreichold:
>  Replace access to type state via raw pointers by using indices and therefore changing the type of Archetype::state

@Ralith Concerning the above, how do you feel about adding a dependency on the [`indexmap`](https://crates.io/crates/indexmap) crate? Using this for `Archetype::state` would support access by cached index as well as hash table look-ups via type ID. Otherwise, I would say I go for a sorted slice in this PR and we consider the previously discussed adaptive approach as a follow-up.

Ralith:
SGTM.

adamreichold:
> except in the mut_cached case which was fastest using the hash table.

What is strange about this is that in the cached case, the access method (hash table, binary search, linear search) does not really enter  the picture(all look-ups are done when the query is first prepared), and so these effects must due to the dereferencing method, i.e. raw pointers versus checked and unchecked indexing. This would also align with unchecked indexing and pointer dereferencing being really close so that we are possibly measuring the cost of offset `Archetype::state` by the index versus following `*const Archetype` directly or at least its impact on the surrounding codegen.

Ralith:
Sorry this has taken me a bit to get back to; I should be able to make a proper pass this weekend at the latest.

src/query.rs
```diff
@@ -89,14 +103,18 @@ unsafe impl<'a, T: Component> Fetch<'a> for FetchRead<T> {
...
-    fn borrow(archetype: &Archetype) {
-        archetype.borrow::<T>();
+    fn borrow(archetype: &Archetype, state: Self::State) {
+        archetype.borrow::<T>(state);
     }
-    fn new(archetype: &'a Archetype) -> Option<Self> {
-        archetype.get_base::<T>().map(Self)
+    #[allow(clippy::needless_question_mark)]
+    fn prepare(archetype: &Archetype) -> Option<Self::State> {
+        Some(archetype.get_state::<T>()?)
```
Ralith, line 111: Why violate `clippy::needless_question_mark`?

src/query.rs
```diff
@@ -352,6 +395,15 @@ impl<'w, Q: Query> QueryBorrow<'w, Q> {
         unsafe { BatchedIter::new(self.meta, self.archetypes.iter(), batch_size) }
     }
 
+    /// Prepare this query for repeated execution
+    ///
+    /// A prepared query can be stored independently of the [world][World] and
+    /// is used to amortise the cost of the looking up the machting [archetypes][Archetype]
+    /// and the relevant type states therein.
```
Ralith, line 402: This explanation is a bit lower-level than I'd like to ask users to understand. How about something like:

```suggestion
    /// A prepared query can be stored independently of the [`World`] to amortize query set-up costs.
```

src/query.rs
```diff
@@ -352,6 +395,15 @@ impl<'w, Q: Query> QueryBorrow<'w, Q> {
...
+    pub fn prepare(self, world: &World) -> PreparedQuery<Q> {
```
Ralith, line 403: Having to pass in the same world twice in `world.query().prepare(&world)` is awkward. Since `PreparedQuery` has to support detecting and rebuilding stale state anyway, why don't we drop the `world` argument and instead construct the `PreparedQuery` in an initially stale state, so that it initializes itself on first use?

src/query.rs
```diff
@@ -762,6 +828,195 @@ macro_rules! tuple_impl {
...
+            state,
+            _query: PhantomData,
+        }
+    }
+
+    /// Borrow the query and the [world][World] as a prerequisite for execution
+    ///
+    /// This will panic if it would violate an existing unique reference
+    /// or construct an invalid unique reference.
+    pub fn borrow<'q>(&'q mut self, world: &'q World) -> PreparedQueryBorrow<'q, Q> {
```
Ralith, line 860: Let's use terminology consistent with non-prepared queries:
```suggestion
    /// Query `world`, using dynamic borrow checking
    ///
    /// This will panic if it would violate an existing unique reference
    /// or construct an invalid unique reference.
    pub fn query<'q>(&'q mut self, world: &'q World) -> PreparedQueryBorrow<'q, Q> {
```

src/query.rs
```diff
@@ -762,6 +828,195 @@ macro_rules! tuple_impl {
...
+        let archetypes = world.archetypes_inner();
+
+        PreparedQueryBorrow::new(meta, archetypes, &*self.state)
+    }
+
+    /// Execute the query using exclusive access to the [world][World]
+    ///
+    /// This avoids the cost of the dynamic borrow checking.
+    /// that is required by [`borrow`][Self::borrow]
+    pub fn iter_mut<'q>(&'q mut self, world: &'q mut World) -> PreparedQueryIter<'q, Q> {
```
Ralith, line 875:
```suggestion
    /// Query a uniquely borrowed world
    ///
    /// This avoids the cost of the dynamic borrow checking.
    /// that is required by [`borrow`][Self::borrow]
    pub fn query_mut<'q>(&'q mut self, world: &'q mut World) -> PreparedQueryIter<'q, Q> {
```

src/query.rs
```diff
@@ -762,6 +828,195 @@ macro_rules! tuple_impl {
...
+        let archetypes = world.archetypes_inner();
+
+        let state: &'q [(usize, <Q::Fetch as Fetch<'q>>::State)] =
+            unsafe { mem::transmute(&*self.state) };
+
+        unsafe { PreparedQueryIter::new(meta, archetypes, state.iter()) }
+    }
+}
+
+/// Represents a combined borrow of both [prepared query][PreparedQuery] and [world][World]
```
Ralith, line 892:
```suggestion
/// Combined borrow of a [`PreparedQuery`] and a [`World`]
```

src/query.rs
```diff
@@ -762,6 +828,195 @@ macro_rules! tuple_impl {
...
+
+impl<Q: Query> Drop for PreparedQueryBorrow<'_, Q> {
+    fn drop(&mut self) {
+        for (idx, state) in self.state {
+            Q::Fetch::release(&self.archetypes[*idx], *state);
+        }
+    }
+}
+
+/// Used to iterate of all entities matching a [prepared query][PreparedQuery]
```
Ralith, line 943:
```suggestion
/// Iterates over all entities matching a [`PreparedQuery`]
```

src/query.rs
```diff
@@ -762,6 +828,195 @@ macro_rules! tuple_impl {
...
+
+    /// Execute the prepared query
+    ///
+    /// Must be called only once per prepared query.
+    // The lifetime narrowing here is required for soundness.
+    pub fn iter<'i>(&'i mut self) -> PreparedQueryIter<'i, Q> {
+        if self.active {
+            panic!("borrow already active");
+        }
+        self.active = true;
```
Ralith, line 926: Since we're not doing the deferred borrow thing that immediate queries use, could we relax this restriction?

src/world.rs
```diff
@@ -47,15 +48,21 @@ pub struct World {
     archetypes: ArchetypeSet,
     /// Maps statically-typed bundle types to archetypes
     bundle_to_archetype: TypeIdMap<u32>,
+    id: usize,
 }
 
 impl World {
     /// Create an empty world
     pub fn new() -> Self {
+        static ID: AtomicUsize = AtomicUsize::new(0);
```
Ralith, line 57: Should we use `AtomicU64` here for insurance against long-running 32-bit systems?

src/query.rs
```diff
@@ -30,18 +32,28 @@ pub unsafe trait Fetch<'a>: Sized {
...
     fn access(archetype: &Archetype) -> Option<Access>;
 
     /// Acquire dynamic borrows from `archetype`
-    fn borrow(archetype: &Archetype);
+    fn borrow(archetype: &Archetype, state: Self::State);
     /// Construct a `Fetch` for `archetype` if it should be traversed
-    fn new(archetype: &'a Archetype) -> Option<Self>;
+    fn new(archetype: &'a Archetype) -> Option<Self> {
+        Self::prepare(archetype).map(|state| Self::execute(archetype, state))
+    }
```
Ralith, line 50: It looks like you've mostly refactored uses of this into the new form. Can it be removed entirely?

Ralith (commented):
Made a full pass. This basically looks great; mostly cosmetic nits and a few minor issues.

src/query.rs
adamreichold: I have added another commit that open codes this instead. Please have a look if you prefer the result and I will squash it. Otherwise, I can drop the commit.

src/world.rs
adamreichold: Changed it to `AtomicU64` so that 32-bit and 64-bit system have the same head room.

src/archetype.rs
```diff
@@ -464,6 +471,36 @@ impl Hasher for TypeIdHasher {
...
+
+    fn get(&self, id: &TypeId) -> Option<&V> {
+        self.search(id).map(move |idx| &self.0[idx].1)
+    }
+
+    fn get_mut(&mut self, id: &TypeId) -> Option<&mut V> {
+        self.search(id).map(move |idx| &mut self.0[idx].1)
+    }
+
+    fn get_index(&self, idx: usize) -> &(TypeId, V) {
```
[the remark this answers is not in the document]
adamreichold: The name is borrowed from `IndexMap`. I think implementing the `Index` trait isn't good as these two methods of indexing into this map are equally natural. Hence I'll rename the method to `get_from_index`.

src/query.rs
adamreichold: Yes, I think you are right. As the lifetime of the returned references is bound to the borrow on `PreparedQueryBorrow` and not on `PreparedQuery` and `World`, I cannot create aliasing mutable references by iterating the same borrow twice.

However, what I do not understand now is why a second call to `QueryBorrow::borrow` is a panic instead of a no-op as `QueryBorrow::iter{,_batched}` have the same lifetime narrowing? (And the transform of an already borrowed query via `with/without` will always release fewer and never more borrows, i.e. it will only leak borrows which I can achieve by forgetting the whole `QueryBorrow` anyway.)

Ralith: I think it would be sound to modify `QueryBorrow` to no-op, yeah. No need to address that in the scope of this PR though.

src/archetype.rs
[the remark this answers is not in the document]
Ralith: `Index` can be implemented for multiple key types, but this is fine too.

src/query.rs
Ralith: Net reduction in LoC, one less trait method, LGTM.

adamreichold:
Sorry, I messed up removing the reactivation test during the last round of rebasing.

adamreichold:
Oh, I just saw 7 more hidden comments...

src/query.rs
adamreichold: Ah, this was a false-positive in an intermediate state (because there was the decay of a reference into a raw pointer between that question mark and `Some`), but since `Self::State` is now `usize` this should go away.

src/query.rs
adamreichold: Will change this, I just need to make sure that there exists a `memo` value that now valid `World` will have, probably by making the ID start at 1 instead of 0.

adamreichold: Added this as a last commit. I did actually go a bit further and removed the `QueryBorrow::prepare` method and just gave `PreparedQuery` a `Default` implementation. Where I have applied this, we ended up storing the queries together with other system-internal state and often ended up initialising them using `Default` so I think this would be the most useful variant.

Ralith: Good idea!

adamreichold:
@Ralith The CI "failure" here seems to be the Googlebot being confusion by the `Co-Authored-By` tag I committed when integrating your suggestions. Should I remove that tag or can it be convinced that you consent to having your suggestions become part of your project?

Ralith:
@googlebot I consent.

Ralith:
Yeah, that was pretty weird. Hopefully it sticks :D

src/query.rs
```diff
@@ -762,6 +817,191 @@ macro_rules! tuple_impl {
...
+            // This memo will not match any world as the first ID will 1.
+            memo: (0, 0),
+            state: Default::default(),
+        }
+    }
+}
+
+impl<Q: Query> PreparedQuery<Q> {
+    #[cold]
+    fn new(world: &World) -> Self {
```
Ralith, line 838: Is this dead code? I didn't see any calls.

src/query.rs
```diff
@@ -762,6 +817,191 @@ macro_rules! tuple_impl {
...
 smaller_tuples_too!(tuple_impl, O, N, M, L, K, J, I, H, G, F, E, D, C, B, A);
 
+/// A prepared query can be stored independently of the [`World`] to amortize query set-up costs.
+pub struct PreparedQuery<Q: Query> {
+    memo: (u64, u64),
+    state: Box<[(usize, <Q::Fetch as Fetch<'static>>::State)]>,
+}
+
+impl<Q: Query> Default for PreparedQuery<Q> {
+    fn default() -> Self {
```
Ralith, line 827: If this is the normal way for users to construct a `PreparedQuery`, we should also expose a `pub fn new() -> Self` for greater discoverability/concision.

Ralith (changes requested):
Thanks, almost there!

src/query.rs
adamreichold: It is called in `PreparedQuery::query` and `::query_mut`:
```rust
if self.memo != world.memo() {
    *self = Self::new(world);
}
```
but I will rename it to `prepare` and expose a public `new` that is equivalent to the `Default` impl.

Ralith: Ah, that's why my grep didn't find it. SGTM.

Ralith (approved):
Looks great, thanks!

PR #160, opened 2021-05-03
RFC: Replace hash table used in archetype by sorted slice

adamreichold:
Measuring the cost of computing the per-component data pointers into an archetype [here](https://github.com/Ralith/hecs/pull/159#issuecomment-831086870), made me consider whether a hash table is the best data structure for storing the offsets and borrows. Assuming that an archetype contains a small number of types, a sorted slice might achieve better cache locality and thereby better performance. This does align with the benchmarks, i.e.
```
test build            ... bench:         209 ns/iter (+/- 4)
test insert           ... bench:     150,688 ns/iter (+/- 5,673)
test iterate_100k     ... bench:      95,986 ns/iter (+/- 1,263)
test iterate_mut_100k ... bench:      71,975 ns/iter (+/- 630)
test remove           ... bench:     164,481 ns/iter (+/- 2,610)
test spawn_batch      ... bench:      63,622 ns/iter (+/- 1,232)
test spawn_static     ... bench:          78 ns/iter (+/- 2)
test spawn_tuple      ... bench:          65 ns/iter (+/- 1)
```
on master and
```
test build            ... bench:         209 ns/iter (+/- 5)
test insert           ... bench:     122,412 ns/iter (+/- 6,853)
test iterate_100k     ... bench:      72,256 ns/iter (+/- 3,885)
test iterate_mut_100k ... bench:      95,857 ns/iter (+/- 1,063)
test remove           ... bench:     129,031 ns/iter (+/- 1,567)
test spawn_batch      ... bench:      45,440 ns/iter (+/- 577)
test spawn_static     ... bench:          68 ns/iter (+/- 1)
test spawn_tuple      ... bench:          67 ns/iter (+/- 2)
```
after this change. But I would also expect the benchmarks being tilted towards small archetypes compared to realistic use cases so I am very much not sure what to make of this...

### DIFF
--- a/src/archetype.rs
+++ b/src/archetype.rs
@@ -27,7 +27,8 @@ use crate::{align, Access, Component, Query};
 /// [`World`](crate::World).
 pub struct Archetype {
     types: Box<[TypeInfo]>,
-    state: TypeIdMap<TypeState>,
+    offsets: SmallTypeMap<usize>,
+    borrows: SmallTypeMap<AtomicBorrow>,
     len: u32,
     entities: Box<[u32]>,
     // UnsafeCell allows unique references into `data` to be constructed while shared references
@@ -62,7 +63,8 @@ impl Archetype {
         let max_align = types.first().map_or(1, |ty| ty.layout.align());
         Self::assert_type_info(&types);
         Self {
-            state: types.iter().map(|ty| (ty.id, TypeState::new(0))).collect(),
+            offsets: SmallTypeMap::new(types.iter().map(|ty| (ty.id, 0))),
+            borrows: SmallTypeMap::new(types.iter().map(|ty| (ty.id, AtomicBorrow::new()))),
             types: types.into(),
             entities: Box::new([]),
             len: 0,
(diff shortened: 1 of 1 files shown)


Ralith:
Interesting idea! The current benchmarks all use one or two components; to fairly judge the impact of this, we should add coverage for a variety of other cases; maybe 5, 10, 20, and 40 component versions of insert/remove/iterate?

Any idea why `iterate_mut_100k` got slower? Does that happen consistently?

Especially for small archetypes, I suspect this could be further improved by replacing the binary search with a linear scan.

adamreichold:
>  to fairly judge the impact of this, we should add coverage for a variety of other cases; maybe 5, 10, 20, and 40 component versions of insert/remove/iterate?

Will do.

> Any idea why iterate_mut_100k got slower? Does that happen consistently?

No, and it does seem to vary from build to build, but it seems to be stable for any given binary. I fear this is bumping against codegen brittleness. I have seen it the other way around on `master` and this branch as well and sometimes both benchmarks end up at approximately 72 µs. (I have also tried adding `black_box` invocations, disabling frequency boost on my CPU and switching to the `performance` governor. But it seems to be due to the produced binary, not the system environment.)

> Especially for small archetypes, I suspect this could be further improved by replacing the binary search with a linear scan.

I am also considering an adaptive approach, i.e. a `TypeIdMap` that internally choose an unsorted list, a sorted list and a hashtable depending on the number of entries. But I fear that the additional branches and reduced code density might reduce the gains. It would also increase code complexity somewhat.

Ralith:
> No, and it does seem to vary from build to build, but it seems to be stable for any given binary. I fear this is bumping against codegen brittleness.

Touchy codegen has definitely been an issue in the past (e.g. https://github.com/Ralith/hecs/pull/103, 808aa45d6d885363f340fff6bb6c1799ad4e6463, others). Sprinkling `#[inline]` around sometimes helps. If it varies between builds of essentially the same code, maybe it's a case of codegen-units interfering with automatic inlining? Maybe codegen-units=1 and/or fat LTO would help.

If both master and this branch have the same transient issue, then presumably these changes aren't responsible, but it'd be nice to mitigate it so we can identify any impact that these changes *do* have. That said, iteration hits this code path once when processing N entities whereas insert/remove hit it per-entity, so the clear win for those is more significant anyway.

> I am also considering an adaptive approach

That occurred to me as well. I don't mind a modest increase in code complexity if it's strongly encapsulated and doesn't introduce hard correctness questions or compromise compile time, which I expect should be the case here. Whether a significant perf benefit can be had is another matter, of course.

We could also explore reducing type metadata lookup regardless of underlying structure, e.g. by copying everything we need into the insert/remove/spawn caches and establishing something similar for queries.

adamreichold:
Closing as this will be subsumed by #159 one way or another.

PR #162, opened 2021-05-09
Fix query_mut permitting unsound double-borrows

### LINKED ISSUE #161: Unsound component access using mutable queries
The following test case shows how mutable queries can be used to alias the same component value via a shared reference and a mutable reference:
```
#[test]
fn unsound_component_access_using_mutable_queries() {
    let mut world = World::new();
    world.spawn((23,));
    world.spawn((42,));

    for (_, (i, m)) in world.query_mut::<(&i32, &mut i32)>() {
        let o = *i;
        *m = 0;
        let n = *i;
        assert_eq!(o, n);
    }
}
```

I think checking the internal consistency of a `Fetch` instance is required to avoid this. Since the relevant information is known at compile time, I also think that this should be possible without runtime overhead, probably via the tuple fetch combinator and the derive macro.

Ralith:
Fixes #161.

Not a huge fan of how this complicates `Fetch`, but it fixes the unsoundness and has no overhead in practice.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -5,6 +5,7 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
+use core::any::TypeId;
 use core::marker::PhantomData;
 use core::ptr::NonNull;
 use core::slice::Iter as SliceIter;
@@ -41,6 +42,9 @@ pub unsafe trait Fetch<'a>: Sized {
     fn new(archetype: &'a Archetype) -> Option<Self>;
     /// Release dynamic borrows acquired by `borrow`
     fn release(archetype: &Archetype);
+
+    /// Invoke `f` for every component type that may be borrowed and whether the borrow is unique
+    fn for_each_borrow(f: impl FnMut(TypeId, bool));
 
     /// Access the `n`th item in this archetype without bounds checking
     ///
@@ -95,6 +99,10 @@ unsafe impl<'a, T: Component> Fetch<'a> for FetchRead<T> {
         archetype.release::<T>();
     }
 
+    fn for_each_borrow(mut f: impl FnMut(TypeId, bool)) {
+        f(TypeId::of::<T>(), false);
+    }
+
     unsafe fn get(&self, n: usize) -> Self::Item {
         &*self.0.as_ptr().add(n)
     }
@@ -132,6 +140,10 @@ unsafe impl<'a, T: Component> Fetch<'a> for FetchWrite<T> {
         archetype.release_mut::<T>();
     }
 
+    fn for_each_borrow(mut f: impl FnMut(TypeId, bool)) {
+        f(TypeId::of::<T>(), true);
+    }
+
     unsafe fn get(&self, n: usize) -> Self::Item {
         &mut *self.0.as_ptr().add(n)
     }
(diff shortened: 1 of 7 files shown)


Ralith:
cc @adamreichold

src/query.rs
```diff
@@ -508,6 +533,17 @@ pub struct QueryMut<'q, Q: Query> {
 
 impl<'q, Q: Query> QueryMut<'q, Q> {
     pub(crate) fn new(meta: &'q [EntityMeta], archetypes: &'q mut [Archetype]) -> Self {
+        // This looks like an ugly O(n^2) loop, but everything's constant after inlining, so in
+        // practice LLVM optimizes it out entirely.
+        Q::Fetch::for_each_borrow(|a, i, unique| {
+            if unique {
```
adamreichold, line 539: One I think I did wonder is whether "properly" checking borrowing is necessary or whether just checking uniqueness of the component types would suffice. Borrowing the same component type twice immutably is a bit useless, but it is also possible using the dynamically checked variant. So it would be a strange API difference, but it would also simplify the implementation by avoid the uniqueness tag and the counter.

src/query.rs
```diff
@@ -701,6 +737,15 @@ macro_rules! tuple_impl {
                 $($name::release(archetype);)*
             }
 
+            #[allow(unused_variables, unused_mut)]
+            fn for_each_borrow(mut f: impl FnMut(TypeId, usize, bool)) {
+                let mut counter = 0;
```
adamreichold, line 742: Coming the borrow checking versus checking uniqueness below, wouldn't that counter need to be used for the whole type hierarchy? I think this way, we'll get examples like
```
((&i32, &mut i32), (&i32, &mut i32)) => ((0,1), (0,1))
```
instead of
```
((&u64, &mut i32), (&u64, &mut i32)) => ((0,1), (2,3))
```
and the above my actually check because `&mut i32` is positioned at `1` in both instances?

src/query.rs
Ralith: How would we avoid the counter in that case?

src/query.rs
Ralith: The callback is invoked for each individual borrow, and not for intermediate constructions like tuples. When tuples are nested, you end up with multiple counters, but an outer counter always ignores the output from the inner; you can see below that we discard the counter value passed in by the inner type. As a result, we should always get the second case.

src/query.rs
adamreichold: > it would also simplify the implementation by avoiding the uniqueness tag and the counter.

Actually, this would not be the case as we still need to avoid checking a component type against itself. But so we do need to resolve https://github.com/Ralith/hecs/pull/162#discussion_r628839322

src/query.rs
adamreichold: Maybe it would be simplest to not do any counting within `for_each_borrow` at all just keep the `i` and `j` counters as mutable variables on the stack of `QueryMut::new`? If `Q::Fetch::for_each_borrow` is guaranteed to always produce the same visitation order, keeping count could be done outside of it entirely.

src/query.rs
adamreichold: We cross posted here...

src/query.rs
adamreichold: > but an outer counter always ignores the output from the inner; you can see below that we discard the counter value passed in by the inner type.

Yes, now that you I mention it, I do see it. I would still argue that keeping the counters on the outer most layer in `QueryMut::new` would actually be the simplest way as it would avoid having to ignore any layer of counting.

Ralith: Good catch; fixed.

adamreichold:
LGTM

Ralith:
Thanks!

Issue #164, opened 2021-05-15
Getting an Entity from an EntityRef


LPGhatguy:
It'd be great to have a method on `EntityRef` that yields the entity that it came from.

Though it's possible to pass an `Entity` along with its corresponding `EntityRef`, it feels redundant!

I'm writing collision handling code that looks roughly like this:

```rust
let mut handle_collision = |projectile, hit| {
    let bullet = projectile.get::<Bullet>()?;

    // Proposed use: Bullet.owner is of type Entity
    if bullet.owner == hit.entity() {
        return None;
    }

    // snip

    None
};

for (a, b) in game.physics.collisions() {
    let a = game.world.entity(*a);
    let b = game.world.entity(*b);

    let _ = handle_collision(a, b);
    let _ = handle_collision(b, a);
}
```

PR #165, opened 2021-05-15
EntityRef::entity accessor for recovering the entity's handle

Ralith:
Fixes https://github.com/Ralith/hecs/issues/164

### DIFF
--- a/src/entity_ref.rs
+++ b/src/entity_ref.rs
@@ -3,18 +3,32 @@ use core::ops::{Deref, DerefMut};
 use core::ptr::NonNull;
 
 use crate::archetype::Archetype;
-use crate::{Component, MissingComponent, Query, QueryOne};
+use crate::{Component, Entity, MissingComponent, Query, QueryOne};
 
 /// Handle to an entity with any component types
 #[derive(Copy, Clone)]
 pub struct EntityRef<'a> {
     archetype: &'a Archetype,
+    generation: u32,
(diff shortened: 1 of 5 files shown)


src/entity_ref.rs
```diff
@@ -3,18 +3,32 @@ use core::ops::{Deref, DerefMut};
...
 
 use crate::archetype::Archetype;
-use crate::{Component, MissingComponent, Query, QueryOne};
+use crate::{Component, Entity, MissingComponent, Query, QueryOne};
 
 /// Handle to an entity with any component types
 #[derive(Copy, Clone)]
 pub struct EntityRef<'a> {
     archetype: &'a Archetype,
+    generation: u32,
```
adamreichold, line 12: Since this borrows `&'a Archetype`, I guess it could also borrow '&'a [EntityMetadata]` instead of storing the generation? So resolving the generation lazily making `world::Iter::next` faster but `EnityRef::entity` slower? Not sure whether this a better trade off though and whether this is measurable at all.

LPGhatguy:
Integrated this branch into our game, I like these changes.

src/entity_ref.rs
Ralith: Yeah, I dunno. I think I'll favor simplicity unless someone cares enough to profile.

Ralith:
Thanks for the feedback, all!

PR #167, opened 2021-05-16
Allow deriving more than one iterator from a query borrow

adamreichold:
There should be no soundness issue here as the borrows can only be narrowed by transforming a query after having iterated it and the lifetime narrowing on the iterator itself will prevent keeping the references it yields around until a second iterator can be derived.

The test case `alias` was indeed not really testing aliasing as it never used the references stored in `_a` and `_b` with overlapping lifetimes.


Ralith:
Thanks!

PR #176, opened 2021-07-24
Apply exponential resize strategy to reserve 

### LINKED ISSUE #175: `spawn_batch` reserves exactly the memory it needs no matter how small leading to constant re-allocation when used repeatedly
Using `spawn_batch` with 5 items 20,000 times takes ~5 seconds and most time is spent copying data to the reallocated archetype storage:
image
[flamegraph.svg.zip](https://github.com/Ralith/hecs/files/6871429/flamegraph.svg.zip)

`spawn_batch` calls `reserve_inner` with the exact amount of items needed https://github.com/Ralith/hecs/blob/4088845dfe572c9e3534f9f818487685a17b6599/src/world.rs#L187
which then calls reserve on the archetype
https://github.com/Ralith/hecs/blob/4088845dfe572c9e3534f9f818487685a17b6599/src/world.rs#L312
https://github.com/Ralith/hecs/blob/4088845dfe572c9e3534f9f818487685a17b6599/src/archetype.rs#L238-L242

Meanwhile the `Archetype::allocate` function used by `World::spawn` bumps the size by 64 if it reaches the max capacity.
https://github.com/Ralith/hecs/blob/4088845dfe572c9e3534f9f818487685a17b6599/src/archetype.rs#L223-L227

One solution to improve performance here would be to adopt this `64` value as the minimum that `spawn_batch` reserves.

However, I also wonder if it would be better to consider a more aggressive growth strategy like that used by `Vec` to minimize the re-allocation and copying needed.

Also, I'm not sure what has already been considered and/or done here or what the limitations would be in this context, but another idea that could help is breaking the archetype into large chunks to prevent the need for copying everything when growing past the current capacity.

Ralith:
Fixes #175.

### DIFF
--- a/src/archetype.rs
+++ b/src/archetype.rs
@@ -222,7 +222,7 @@ impl Archetype {
     /// Every type must be written immediately after this call
     pub(crate) unsafe fn allocate(&mut self, id: u32) -> u32 {
         if self.len as usize == self.entities.len() {
-            self.grow(self.len.max(64));
+            self.grow(64);
         }
 
         self.entities[self.len as usize] = id;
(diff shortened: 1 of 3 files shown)


Ralith:
Test failure is spurious, due to a cosmetic change in error formatting for the trybuild tests in Rust beta.

PR #186, opened 2021-09-02
Reexport QueryOneError

t-mw:
Exports the  error type for `query_one_mut`. The error type for `query_one` is already exported.


t-mw:
@googlebot I signed it!

Ralith:
Thanks!

Issue #172, opened 2021-07-06
Use NonZeroU32 for entity generation.


zakarumych:
This is possible if generations start with 1. And this makes `Option<Entity>` and `Entity` sizes match.

louisgjohnson:
I was just browsing the code to see how much work would be required for this, what happens if the generation starts with zero, should the generation value be None? So the generation typing should be `Option<NonZeroU32>`

for example the EntityMeta has a const value called Empty that returns 0 for the generation

zakarumych:
The generation should never be 0.

Ralith:
Implemented in #190.

PR #190, opened 2021-09-10
Use NonZeroU32 for entity generation

ouz-a:
As regards to issue: https://github.com/Ralith/hecs/issues/172


ouz-a:
@googlebot I signed it!

src/entities.rs
```diff
@@ -29,7 +30,7 @@ impl Entity {
     /// `World::spawn_at` for easy serialization. Alternatively, consider `id` for more compact
     /// representation.
     pub fn to_bits(self) -> u64 {
-        u64::from(self.generation) << 32 | u64::from(self.id)
+        NonZeroU64::from(self.generation).get() << 32 | u64::from(self.id)
```
Ralith, line 33:
```suggestion
        u64::from(self.generation.get()) << 32 | u64::from(self.id)
```

src/entities.rs
```diff
@@ -38,7 +39,7 @@ impl Entity {
     /// `World::spawn_at` for easy serialization.
     pub fn from_bits(bits: u64) -> Self {
         Self {
-            generation: (bits >> 32) as u32,
+            generation: Entities::ret_non_zero_u32((bits >> 32) as u32),
```
Ralith, line 42: This is unsound, because `bits >> 32` might be zero. We should use the safe, checked constructor for `NonZeroU32` and modify `from_bits` to return `Option`.

src/entities.rs
```diff
@@ -464,7 +477,12 @@ impl Entities {
             init(id, &mut self.meta[id as usize].location);
         }
     }
-
+    // Since nonzero ops are not const yet (https://github.com/rust-lang/rust/issues/84186)
+    // unsafe is required to bypass the compiler error
+    /// Returns `NonZeroU32` from given number
+    pub(crate) const fn ret_non_zero_u32(num: u32) -> NonZeroU32 {
+        unsafe { NonZeroU32::new_unchecked(num) }
+    }
```
Ralith, line 485: This should be dropped in favor of invoking `NonZeroU32::new_unchecked` directly; it's not saving a significant amount of verbosity and it is a major safety hazard.

src/entities.rs
```diff
@@ -340,7 +349,8 @@ impl Entities {
         if meta.generation != entity.generation {
             return Err(NoSuchEntity);
         }
-        meta.generation += 1;
+        let to_add = u32::from(meta.generation) + 1;
+        meta.generation = Entities::ret_non_zero_u32(to_add);
```
Ralith, line 353: This is unsound, because `u32::from(meta.generation) + 1` might wrap. Use `checked_add` and handle wrapping explicitly.

Ralith (changes requested):
Thanks for working on this! A lighter weight `Option<Entity>` will be very cool to have.

entities.rs
```diff
@@ -0,0 +1,753 @@
+use alloc::vec::Vec;
```
Ralith, line 1: You seem to have added a copy of the entire file to the root.

src/entities.rs
```diff
@@ -29,17 +31,21 @@ impl Entity {
...
-            id: bits as u32,
+    pub fn from_bits(bits: u64) -> Option<Self> {
+        if NonZeroU32::new((bits >> 32) as u32).is_some() {
+            Some(Self {
+                generation: NonZeroU32::new((bits >> 32) as u32).unwrap(),
+                id: bits as u32,
+            })
+        } else {
+            None
         }
```
Ralith, line 49: It's redundant to call `new` twice.
```suggestion
        Some(Self {
            generation: NonZeroU32::new((bits >> 32) as u32)?
            id: bits as u32,
        })
```
You could also use `map`.

src/entities.rs
```diff
@@ -78,7 +84,15 @@ impl<'de> serde::Deserialize<'de> for Entity {
         D: serde::Deserializer<'de>,
     {
         let bits = u64::deserialize(deserializer)?;
-        Ok(Entity::from_bits(bits))
+
+        if Entity::from_bits(bits).is_some() {
+            Ok(Entity::from_bits(bits).unwrap())
```
Ralith, line 89: `from_bits` should not be called twice. Use `match`, or `.ok_or_else(|| error_goes_here)?`.

src/entities.rs
```diff
@@ -78,7 +84,15 @@ impl<'de> serde::Deserialize<'de> for Entity {
...
+        } else {
+            Err(serde::de::Error::invalid_value(
+                serde::de::Unexpected::Unsigned(bits),
+                &"Expected unsigned value",
```
Ralith, line 93: This error message doesn't make sense.

src/entities.rs
Ralith: Still an issue.

src/entities.rs
```diff
@@ -29,17 +31,21 @@ impl Entity {
...
     /// representation.
     pub fn to_bits(self) -> u64 {
-        u64::from(self.generation) << 32 | u64::from(self.id)
+        NonZeroU64::from(self.generation).get() << 32 | u64::from(self.id)
     }
 
     /// Reconstruct an `Entity` previously destructured with `to_bits`
     ///
     /// Useful for storing entity IDs externally, or in conjunction with `Entity::to_bits` and
     /// `World::spawn_at` for easy serialization.
```
Ralith, line 40: We should document the conditions in which `None` is returned, i.e. when `bits` does not correspond to a possible return value from `to_bits`.

Ralith (changes requested):

src/entities.rs
```diff
@@ -340,7 +362,8 @@ impl Entities {
         if meta.generation != entity.generation {
             return Err(NoSuchEntity);
         }
-        meta.generation += 1;
+        meta.generation = NonZeroU32::new(u32::from(meta.generation).wrapping_add(1))
+            .unwrap_or_else(|| panic!("Error freeing entity Entity Generation overflows"));
```
Ralith, line 366: Instead of panicing, return `NonZeroU32::new(1).unwrap()`.

entities.rs
ouz-a: ups

src/entities.rs
```diff
@@ -27,26 +27,26 @@ impl Entity {
...
     pub fn from_bits(bits: u64) -> Option<Self> {
-        if NonZeroU32::new((bits >> 32) as u32).is_some() {
-            Some(Self {
-                generation: NonZeroU32::new((bits >> 32) as u32).unwrap(),
-                id: bits as u32,
-            })
-        } else {
-            None
-        }
+        NonZeroU32::new((bits >> 32) as u32).map(|_| Self {
```
ouz-a, line 46: Now that to_bits returns Option<NonZeroU64> shouldn't we assume that the user will deal with the option, and just use new_unchecked here ?

src/serialize/column.rs
```diff
@@ -909,7 +909,7 @@ mod tests {
             Token::TupleEnd,
             Token::Tuple { len: 1 },
             Token::Tuple { len: 1 },
-            Token::U64(e2.to_bits()),
+            Token::U64(e2.to_bits().unwrap().into()),
```
ouz-a, line 912: I didn't know how to handle these and assumed it should pass the test

src/entities.rs
```diff
@@ -25,22 +27,26 @@ impl Entity {
...
     /// representation.
-    pub fn to_bits(self) -> u64 {
-        u64::from(self.generation) << 32 | u64::from(self.id)
+    pub fn to_bits(self) -> Option<NonZeroU64> {
+        NonZeroU64::new(NonZeroU64::from(self.generation).get() << 32 | u64::from(self.id))
     }
 
-    /// Reconstruct an `Entity` previously destructured with `to_bits`
+    /// Reconstruct an `Entity` previously destructured with `to_bits` if the `bits` is a value
+    /// that can correspond to possible `NonZeroU64`
```
Ralith, line 40: The new text is incorrect. For example, 1 is a valid `NonZeroU64` but would yield `None` here. How about something like "or `None` for invalid bitpatterns"?

src/entities.rs
```diff
@@ -25,22 +27,26 @@ impl Entity {
     ///
     /// No particular structure is guaranteed for the returned bits.
     ///
+    /// If bits does not correspond to a possible `NonZeroU64` value returns `None`
+    ///
```
Ralith, line 31: This text is confusing, as there is no user-controlled value named "bits", and the function cannot possibly return `None`.

src/entities.rs
```diff
@@ -25,22 +27,26 @@ impl Entity {
...
     /// Useful for storing entity IDs externally, or in conjunction with `Entity::from_bits` and
     /// `World::spawn_at` for easy serialization. Alternatively, consider `id` for more compact
     /// representation.
-    pub fn to_bits(self) -> u64 {
-        u64::from(self.generation) << 32 | u64::from(self.id)
+    pub fn to_bits(self) -> Option<NonZeroU64> {
+        NonZeroU64::new(NonZeroU64::from(self.generation).get() << 32 | u64::from(self.id))
```
Ralith, line 36: This should not return `Option`, because every `Entity` is guaranteed to have a (nonzero) bitpattern. Additionally, there's no point using `NonZeroU64::from` internally only to immediately convert back to a regular `u64`.

```suggestion
    pub fn to_bits(self) -> NonZeroU64 {
        unsafe { NonZeroU64::new_unchecked(u64::from(self.generation.get()) << 32 | u64::from(self.id)) }
```

src/entities.rs
Ralith: Using `new_unchecked` here would be unsound, because the user could pass in 0.

src/entities.rs
```diff
@@ -25,22 +27,26 @@ impl Entity {
...
+    /// 
     ///
     /// Useful for storing entity IDs externally, or in conjunction with `Entity::to_bits` and
     /// `World::spawn_at` for easy serialization.
-    pub fn from_bits(bits: u64) -> Self {
-        Self {
-            generation: (bits >> 32) as u32,
+    pub fn from_bits(bits: u64) -> Option<Self> {
+        NonZeroU32::new((bits >> 32) as u32).map(|_| Self {
+            generation: NonZeroU32::new((bits >> 32) as u32).unwrap(),
```
Ralith, line 47: It is redundant to call `NonZeroU32::new` twice. The argument passed in by `map`, which you're currently throwing away with `_`, is desired `NonZeroU32` value already. You could also use the example I gave in the previous review, which instead uses `?`.

src/entities.rs
```diff
@@ -78,7 +84,14 @@ impl<'de> serde::Deserialize<'de> for Entity {
         D: serde::Deserializer<'de>,
     {
         let bits = u64::deserialize(deserializer)?;
-        Ok(Entity::from_bits(bits))
+
+        match Entity::from_bits(bits) {
+            Some(ent) => Ok(ent),
+            None => Err(serde::de::Error::invalid_value(
+                serde::de::Unexpected::Unsigned(bits),
+                &"`bits` does not correpond to possible `NonZeroU64` value ",
```
Ralith, line 92: This error text is incorrect, because many nonzero `u64`s still have zero for their upper 32 bits, making them invalid. You've also got a couple typos, and a user preforming serde deserialization is not aware of any variable named "bits." Finally, [the docs for `invalid_value`](https://docs.serde.rs/serde/de/trait.Error.html#method.invalid_value) indicate that the second argument should describe the expected value, not elaborate on what was actually found.

```suggestion
                &"a valid `Entity` bitpattern",
```

Ralith (changes requested):

ouz-a:
hopefully, it's all clear now 😄

Ralith (approved):
Thanks!

src/entities.rs
```diff
@@ -3,10 +3,12 @@ use core::cmp;
 use core::convert::TryFrom;
 use core::iter::ExactSizeIterator;
 use core::ops::Range;
+use core::ptr::NonNull;
```
Ralith, line 6: Unnecessary `use`.
```suggestion
```

src/entities.rs
```diff
@@ -3,10 +3,12 @@ use core::cmp;
...
 use core::sync::atomic::{AtomicI64, Ordering};
 use core::{fmt, mem};
 #[cfg(feature = "std")]
 use std::error::Error;
+use std::num::{NonZeroU32, NonZeroU64};
```
Ralith, line 11: no-std support means we have to use a different path for this.
```suggestion
use core::num::{NonZeroU32, NonZeroU64};
```

Ralith (changes requested):
CI found a few last minor things.

PR #194, opened 2021-09-18
Cloneable component specialization of EntityBuilder

### LINKED ISSUE #193: Feature: re-usable component bundles
I'd like to efficiently re-use component bundles that aren't known statically.
Implementing `DynamicBundle` myself seems painful and not intended.
Re-using can only happen if all of the components are `Clone`.
Ideally, we'd have `impl DynamicBundle for &'a EntityBuilder` if we know that all of the components are cloneable. This would allow me to just keep an `EntityBuilder` around and use it as a bundle.

Ralith:
This is convenient for storing dynamic entity prototypes, which may be used to efficiently spawn individual identical entities on a recurring basis.

Fixes #193.

There's one aspect of the current draft I'm not happy with: a double indirect function call per component when building an entity from a cloneable builder. There's no escaping at least one indirect call per component, but it'd be nice not to have two. This is difficult because we need temporary storage for the cloned component value so that we can construct a pointer to it in `DynamicBundle::put` which the caller can move out of. If we refactored `DynamicBundle::put` to take pointers to write into instead, bundles would have to handle random access by `TypeId`, which is awkward and might compromise performance. Maybe we should allocate reusable scratch space on the heap? Probably better benchmark to be sure...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 - `Or` query combinator, allowing a single query to select entities that satisfy at least one of two
   sub-queries.
 - `EntityRef::has` to efficiently check for the presence of a component without borrowing it
+- `EntityBulider::new_cloneable` constructs an `EntityBuilder` that builds a `ReusableBuiltEntity`,
+  which can be spawned repeatedly by reference.
 
 ### Changed
 - Added a niche to `Entity`, making `Option<Entity>` the same size as a bare `Entity`. As a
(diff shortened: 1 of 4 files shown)


Ralith:
Benchmarking indicates that this is more than twice as efficient as repeatedly re-populating an `EntityBuilder`. Between that and the effective encapsulation, I'm inclined to go ahead.

```
test build                          ... bench:         128 ns/iter (+/- 5)
test build_cloneable                ... bench:          49 ns/iter (+/- 6)
```

PR #196, opened 2021-09-27
add dynamic query support from feather-hecs

jazzay:
This brings over the dynamic query changes from a hard fork of hecs:
https://github.com/feather-rs/feather-hecs

This allows us to build and query components dynamically at runtime.


src/dynamic_query.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Returns a pointer to the raw component data.
+    pub fn ptr(&self) -> NonNull<u8> {
+        self.ptr
+    }
+
+    /// Casts this component slice to a slice of type `T`.
+    ///
+    /// # Panics
+    /// Panics if `T` is not the type of the components in this slice.
+    pub fn as_slice<T: 'static>(&self) -> &[T] {
```
Ralith, line 66: This is unsound because `&[T]` may outlive the query (and indeed the entire `World`. `ComponentSlice` should have a lifetime parameter bounded by the lifetime of the query, which should be propagated to this return value.

src/dynamic_query.rs
```diff
@@ -0,0 +1,130 @@
...
+                id,
+            })
+    }
+
+    /// Returns an iterator over pointers to components of the given type
+    /// yielded by this query.
+    pub fn iter_component_slices<'a>(
+        &'a self,
+        component_type: TypeId,
+    ) -> impl Iterator<Item = ComponentSlice> + 'a {
```
Ralith, line 107: This is unsound because no dynamic borrow-checking is performed, and the use of `&self` here and in `World::query_dynamic` allows for other outstanding (potentially mutable) borrows. Either something analogous to `Fetch::borrow` is needed to perform dynamic borrow-checking, or a lot of stuff should be switched to `&mut self` to statically prevent a dynamic query from coexisting with any other component borrow.

src/dynamic_query.rs
```diff
@@ -0,0 +1,130 @@
...
+
+/// A pointer to a slice of component data from a [`DynamicQuery`].
+pub struct ComponentSlice {
+    typ: TypeId,
+    ptr: NonNull<u8>,
+    len: usize,
+    component_layout: Layout,
+}
+
+impl ComponentSlice {
```
Ralith, line 41: There should be a safe way to get mutable access to the `write_types` of the query.

Ralith (changes requested):
Thanks, this looks like a good start on https://github.com/Ralith/hecs/issues/69! There are, however, some soundness issues as-is, and I think we'll want to iterate on the API a bit.

Issue #127, opened 2021-01-08
Command buffer helper


Ralith:
We should supply a mechanism for recording entities to spawn/despawn on a world at some point in the future, to simplify the performant implementation of code that wants to intermingle those operations with iteration over a query. In particular, there's an opportunity to do cheeky `EntityBuilder`-style region allocation and reuse shared between every pending spawn operation, saving some overhead compared to expecting the user to e.g. keep a `Vec<EntityBuilder>`.

Ralith:
Implemented in #199.

PR #199, opened 2021-11-02
command buffer helper

ouz-a:
As regards the [127](https://github.com/Ralith/hecs/issues/127#issue-782330578) this is a simple and efficient solution to the issue.

It's faster than the naive Vec<EntityBuilder> approach.


src/entity_recorder.rs
```diff
@@ -0,0 +1,173 @@
...
+        let len = bundle.type_info().len();
+        unsafe {
+            bundle.put(|ptr, ty| self.add_inner(ptr, ty, ent));
+        }
+        self.ent.push((ent, len));
+        self
+    }
+
+    pub(crate) fn sort_buffered(&mut self) {
+        self.info[..].sort_unstable_by_key(|z| (z.2, z.0));
```
Ralith, line 89: Rather than sorting the entire vector by a compound key, we should sort the individual slices corresponding to each entity by the `TypeInfo` alone. That's less work, and lets us avoid storing an `Entity` per component.

src/entity_recorder.rs
```diff
@@ -0,0 +1,173 @@
...
+/// assert_eq!(world.contains(reserved_ent), true);
+/// ```
+pub struct EntityRecorder {
+    pub(crate) ent: Vec<(Entity, usize)>,
+    storage: NonNull<u8>,
+    layout: Layout,
+    cursor: usize,
+    info: Vec<(TypeInfo, usize, Entity)>,
+    ids: Vec<TypeId>,
+    pub(crate) mark: usize,
```
Ralith, line 35: This should be a local variable in the code that's doing the spawning; there's no need for it to exist outside of that.

src/entity_recorder.rs
```diff
@@ -0,0 +1,173 @@
...
+/// # use hecs::*;
+/// let mut world = World::new();
+/// let reserved_ent = world.reserve_entity();
+/// let mut entity_recorder = EntityRecorder::new();
+/// entity_recorder.record_entity(reserved_ent,(true,0.10));
+/// world.spawn_recorded(&mut entity_recorder); // recorder can now be reused
+/// assert_eq!(world.contains(reserved_ent), true);
+/// ```
+pub struct EntityRecorder {
+    pub(crate) ent: Vec<(Entity, usize)>,
```
Ralith, line 29: Replace this tuple with a private struct with named elements for clarity.

src/entity_recorder.rs
```diff
@@ -0,0 +1,173 @@
...
+    storage: NonNull<u8>,
+    layout: Layout,
+    cursor: usize,
+    info: Vec<(TypeInfo, usize, Entity)>,
```
Ralith, line 33: As above.

src/world.rs
```diff
@@ -135,6 +135,31 @@ impl World {
...
+    /// ```
+    /// # use hecs::*;
+    /// let mut world = World::new();
+    /// let a = world.reserve_entity();
+    /// let mut recorder = EntityRecorder::new();
+    /// recorder.record_entity(a, (false,0.0));
+    /// world.spawn_recorded(&mut recorder);
+    /// assert!(world.contains(a));
+    /// ```
+    pub fn spawn_recorded(&mut self, recorder: &mut EntityRecorder) {
```
Ralith, line 152: This should be a method on `EntityRecorder` that takes `&mut World`, perhaps named `run_on`. That way `ReadyRecorder` and various other details of `EntityRecorder` don't need to be exposed, and `World`'s API is unaffected.

Ralith (changes requested):
Thanks! The CI failure is unrelated, unfortunately that happens every time a beta changes the formatting of the error in question.

Please also include unit tests and a benchmark.

tests/tests.rs
```diff
@@ -272,6 +272,18 @@ fn dynamic_components() {
...
+#[test]
+fn spawn_buffered_entity() {
+    let mut world = World::new();
+    let mut buffer = CommandBuffer::new();
+    let ent = world.reserve_entity();
+
+    buffer.spawn_at(ent, (1, true));
+    buffer.run_on(&mut world);
+
+    assert!(world.contains(ent));
```
Ralith, line 284: This test should verify that the components have the correct values, and should spawn multiple entities having some distinct component types and some of the same type with different values.

src/command_buffer.rs
```diff
@@ -0,0 +1,221 @@
...
+/// Data required to store components and their offset  
+pub struct InfoBuffer {
+    pub ty_info: TypeInfo,
+    pub size: usize,
+}
+
+/// Data of buffered 'entity' and its relative position in component data
+pub struct EntityIndex {
+    pub entity: Entity,
+    pub beg: usize,
```
Ralith, line 219: We can afford a couple additional characters for readability here.
```suggestion
    pub begin: usize,
```

src/command_buffer.rs
```diff
@@ -0,0 +1,221 @@
...
+        let end = self.buffer.ent[0].end - self.buffer.ent[0].beg;
+        for info in self.buffer.info.drain(0..end) {
+            let ptr = self.buffer.storage.as_ptr().add(info.size);
+            f(ptr, info.ty_info);
+        }
+    }
+}
+
+/// Data required to store components and their offset  
+pub struct InfoBuffer {
```
Ralith, line 211: This structure is not a buffer.
```suggestion
pub struct ComponentInfo {
```

src/lib.rs
```diff
@@ -72,6 +73,7 @@ mod world;
 pub use archetype::Archetype;
 pub use batch::{BatchIncomplete, BatchWriter, ColumnBatch, ColumnBatchBuilder, ColumnBatchType};
 pub use bundle::{Bundle, DynamicBundle, MissingComponent};
+pub use command_buffer::{CommandBuffer, ReadyBuffer};
```
Ralith, line 76: `ReadyBuffer` is not needed in the public API.
```suggestion
pub use command_buffer::CommandBuffer;
```

src/command_buffer.rs
```diff
@@ -0,0 +1,221 @@
...
+    pub ty_info: TypeInfo,
+    pub size: usize,
+}
+
+/// Data of buffered 'entity' and its relative position in component data
+pub struct EntityIndex {
+    pub entity: Entity,
+    pub beg: usize,
+    pub end: usize,
+}
```
Ralith, line 221: None of these need to be accessed from outside of this module, and hence they should not be `pub`.

src/command_buffer.rs
```diff
@@ -0,0 +1,221 @@
...
+pub struct InfoBuffer {
+    pub ty_info: TypeInfo,
+    pub size: usize,
+}
+
+/// Data of buffered 'entity' and its relative position in component data
+pub struct EntityIndex {
+    pub entity: Entity,
+    pub beg: usize,
+    pub end: usize,
```
Ralith, line 220: These should have comments describing what they index into.

src/command_buffer.rs
```diff
@@ -0,0 +1,221 @@
...
+            let ptr = self.buffer.storage.as_ptr().add(info.size);
+            f(ptr, info.ty_info);
+        }
+    }
+}
+
+/// Data required to store components and their offset  
+pub struct InfoBuffer {
+    pub ty_info: TypeInfo,
+    pub size: usize,
```
Ralith, line 213: This appears to be the offset of the component in memory, not anything's size.
```suggestion
    /// Position in `storage`
    pub offset: usize,
```

src/command_buffer.rs
```diff
@@ -0,0 +1,221 @@
...
+            end,
+        });
+    }
+
+    fn get_beg(&self) -> usize {
+        match self.ent.len() {
+            0 => 0,
+            _ => self.ent[self.ent.len() - 1].end,
+        }
+    }
```
Ralith, line 101: This is only called once, so it doesn't really merit being a separate helper function. It could also be written more concisely as `self.ent.first().map_or(0, |x| x.end)`.

src/command_buffer.rs
```diff
@@ -0,0 +1,221 @@
...
+/// # use hecs::*;
+/// let mut world = World::new();
+/// let reserved_ent = world.reserve_entity();
+/// let mut entity_buffer = CommandBuffer::new();
+/// entity_buffer.spawn_at(reserved_ent,(true,0.10));
+/// entity_buffer.run_on(&mut world); // buffer can now be reused
+/// assert!(world.contains(reserved_ent));
+/// ```
+pub struct CommandBuffer {
+    pub(crate) ent: Vec<EntityIndex>,
```
Ralith, line 30: This doesn't need to be `pub(crate)`, or abbreviated.

src/command_buffer.rs
```diff
@@ -0,0 +1,221 @@
...
+        match self.ent.len() {
+            0 => 0,
+            _ => self.ent[self.ent.len() - 1].end,
+        }
+    }
+
+    /// Spawn every `entity` recorded with their components
+    ///
+    /// Useful for recording and spawning entities at some point in the future
+    ///
```
Ralith, line 106: This doesn't add any information the reader won't have already encountered by getting this far.
```suggestion
```

src/command_buffer.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// assert!(world.contains(a));
+    /// ```
+    pub fn run_on(&mut self, world: &mut World) {
+        let mut mark: usize = 0;
+        self.sort_buffered();
+        (0..self.ent.len()).for_each(|_| {
+            let (ent, comps) = self.build(mark);
+            world.spawn_at(ent, comps);
+            mark += 1;
+        });
```
Ralith, line 124: This should be a plain for loop. The less-obvious functional style doesn't do us any good here.

src/command_buffer.rs
```diff
@@ -0,0 +1,221 @@
...
+        self.clear();
+    }
+
+    pub(crate) fn sort_buffered(&mut self) {
+        for index in self.ent.iter() {
+            self.info[index.beg..index.end].sort_unstable_by_key(|z| z.ty_info);
+        }
+    }
```
Ralith, line 132: This doesn't need to be a separate helper function; we spend as many lines defining and calling it as its entire body occupies!

src/command_buffer.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    pub(crate) fn build(&mut self, mark: usize) -> (Entity, ReadyBuffer<'_>) {
```
Ralith, line 134: This isn't needed from outside the module.
```suggestion
    fn build(&mut self, mark: usize) -> (Entity, ReadyBuffer<'_>) {
```

src/command_buffer.rs
```diff
@@ -0,0 +1,221 @@
...
+        let end = self.ent[0].end - self.ent[0].beg;
```
Ralith, line 135: Is this an end, or a length?

ouz-a:
> Please respond to comments you've addressed, either marking them as resolved if you think you're done with them, or with a comment if you think additional attention is required.

I didn't realize I had to reply sorry 😅

tests/tests.rs
ouz-a: I hope new test covers it all

src/command_buffer.rs
```diff
@@ -27,12 +27,13 @@ use crate::{align, DynamicBundle};
...
 pub struct CommandBuffer {
-    pub(crate) ent: Vec<EntityIndex>,
+    entity: Vec<EntityIndex>,
     storage: NonNull<u8>,
     layout: Layout,
     cursor: usize,
-    info: Vec<InfoBuffer>,
+    info: Vec<ComponentInfo>,
     ids: Vec<TypeId>,
+    mark: usize,
```
ouz-a, line 36: This is very useful to indicate where we are in regards to a spawned entity, another solution would be to drain Vec as entities get spawned but I think that's not a very beneficial performance wise

If you have other solutions to this, please let me know 😃

tests/tests.rs
```diff
@@ -272,6 +272,34 @@ fn dynamic_components() {
...
+    buffer.spawn_at(ent3, (2 as u8, "qwe", 101.103, false));
+
+    buffer.run_on(&mut world);
+
+    let be_true = world.get::<bool>(ent).unwrap();
+    let be_hecs = world.get::<&str>(ent1).unwrap();
+    let be_false = world.get::<bool>(ent2).unwrap();
+    let be_two = world.get::<u8>(ent3).unwrap();
+
+    assert!(world.contains(ent));
```
Ralith, line 296: This `assert` is now redundant to the `unwrap()` above. As a style nit, I'd also prefer for the `get` calls to be inside the `assert_eq` invocations; this makes changes easier and less error-prone.

src/command_buffer.rs
```diff
@@ -0,0 +1,217 @@
...
+pub struct ComponentInfo {
+    ty_info: TypeInfo,
+    /// Position in 'storage'
+    offset: usize,
+}
+
+/// Data of buffered 'entity' and its relative position in component data
+pub struct EntityIndex {
+    entity: Entity,
+    /// Indexes used to find amount of component entity has in the `ComponentInfo`
```
Ralith, line 214:
```suggestion
    // Range of associated indices in `CommandBuffer`'s `info` member
```

src/command_buffer.rs
```diff
@@ -0,0 +1,217 @@
...
+        for info in self.buffer.info.drain(0..end) {
+            let ptr = self.buffer.storage.as_ptr().add(info.offset);
+            f(ptr, info.ty_info);
+        }
+    }
+}
+
+/// Data required to store components and their offset  
+#[derive(Debug)]
+pub struct ComponentInfo {
```
Ralith, line 205: This does not need to be `pub`.

src/command_buffer.rs
```diff
@@ -0,0 +1,217 @@
...
+    ty_info: TypeInfo,
+    /// Position in 'storage'
+    offset: usize,
+}
+
+/// Data of buffered 'entity' and its relative position in component data
+pub struct EntityIndex {
```
Ralith, line 212: As above.

src/command_buffer.rs
```diff
@@ -0,0 +1,217 @@
...
+/// # use hecs::*;
+/// let mut world = World::new();
+/// let reserved_ent = world.reserve_entity();
+/// let mut entity_buffer = CommandBuffer::new();
+/// entity_buffer.spawn_at(reserved_ent,(true,0.10));
+/// entity_buffer.run_on(&mut world); // buffer can now be reused
+/// assert!(world.contains(reserved_ent));
+/// ```
+pub struct CommandBuffer {
+    entity: Vec<EntityIndex>,
```
Ralith, line 30: `entity` is singular. Because this stores information about multiple entities, it should be a plural, e.g. `entities`.

src/command_buffer.rs
```diff
@@ -0,0 +1,217 @@
...
+        self.info.push(ComponentInfo {
+            ty_info: ty,
+            offset,
+        });
+        self.cursor = end;
+    }
+
+    /// Buffer 'reserved_entity' with 'bundle'
+    ///
+    /// Can also be used with 'EntityBuilder'
```
Ralith, line 82: We don't need to call out the compatibility of specific implementations of `DynamicBundle`, especially when there's no special interaction with them.
```suggestion
```

src/command_buffer.rs
Ralith: There's no reason for this information to be stored persistently inside the `CommandBuffer`; it is only needed inside `fn run_on`, and hence should be a local variable there, passed downward as needed (e.g. into a member of `ReadyBuffer`). Keeping it around in between calls is confusing, wastes storage, and increases the risk of bugs.

src/command_buffer.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn type_info(&self) -> Vec<TypeInfo> {
+        let end =
+            self.buffer.entity[self.buffer.mark].end - self.buffer.entity[self.buffer.mark].begin;
+        self.buffer.info[0..end].iter().map(|x| x.ty_info).collect()
+    }
+
+    unsafe fn put(self, mut f: impl FnMut(*mut u8, TypeInfo)) {
+        let end =
+            self.buffer.entity[self.buffer.mark].end - self.buffer.entity[self.buffer.mark].begin;
+        for info in self.buffer.info.drain(0..end) {
```
Ralith, line 196: Repeatedly removing a prefix of a vector is wasteful, because the entire remaining length must be copied each time, resulting in quadratic overall runtime.

Ralith (changes requested):

src/command_buffer.rs
```diff
@@ -178,40 +183,43 @@ impl Default for CommandBuffer {
...
     }
 
     unsafe fn put(self, mut f: impl FnMut(*mut u8, TypeInfo)) {
-        let end =
-            self.buffer.entity[self.buffer.mark].end - self.buffer.entity[self.buffer.mark].begin;
-        for info in self.buffer.info.drain(0..end) {
+        for info in self
+            .buffer
+            .info
+            .drain(self.buffer.entities[self.mark].begin..self.buffer.entities[self.mark].end)
```
ouz-a, line 204: Hope this improves the performance issue

src/command_buffer.rs
```diff
@@ -107,25 +104,35 @@ impl CommandBuffer {
     /// assert!(world.contains(a));
     /// ```
     pub fn run_on(&mut self, world: &mut World) {
-        for index in self.entity.iter() {
+        let mut mark = self.entities.len() - 1;
```
ouz-a, line 107: now mark goes from last to first entity

src/command_buffer.rs
```diff
@@ -0,0 +1,225 @@
...
+            cursor: 0,
+            info: Vec::new(),
+            ids: Vec::new(),
+        }
+    }
+}
+
+/// The output of an '[CommandBuffer]` suitable for passing to
+/// [`World::spawn_into`](crate::World::spawn_into)
+pub struct ReadyBuffer<'a> {
```
Ralith, line 184: Unnecessary `pub`

src/command_buffer.rs
```diff
@@ -0,0 +1,225 @@
...
+        let addr = self.storage.as_ptr().add(offset);
+        ptr::copy_nonoverlapping(ptr, addr, ty.layout().size());
+        self.info.push(ComponentInfo {
+            ty_info: ty,
+            offset,
+        });
+        self.cursor = end;
+    }
+
+    /// Buffer 'reserved_entity' with 'bundle'
```
Ralith, line 79: "reserved_entity" doesn't appear anywhere else, and the use of "buffer" as a verb here is unconventional. Additionally, you probably meant to use backticks.
```suggestion
    /// Record an entity spawn operation
```

src/command_buffer.rs
```diff
@@ -0,0 +1,225 @@
...
+        let begin = self.entities.last().map_or(0, |x| x.end);
+        let end = begin + len;
+        self.entities.push(EntityIndex {
+            entity: ent,
+            begin,
+            end,
+        });
+    }
+
+    /// Spawn every `entity` recorded with their components
```
Ralith, line 94: Only use backticks to refer to code. There is no variable named `entity` in either the example or the method signature.
```suggestion
    /// Spawn every entity recorded with their components
```

src/command_buffer.rs
```diff
@@ -0,0 +1,225 @@
...
+            }
+        }
+        self.clear();
+    }
+
+    fn build(&mut self, mark: usize) -> (Entity, ReadyBuffer<'_>) {
+        // Amount of components entity has
+        let len = self.entities[mark].end - self.entities[mark].begin;
+        self.ids
+            .extend(self.info[0..len].iter().map(|x| x.ty_info.id()));
```
Ralith, line 127: Why do we always use the first `len` entries here, regardless of which entity is being spawned?

Ralith (changes requested):
Please also resolve the lint.

src/command_buffer.rs
ouz-a: That's a ... relic from the past and shouldn't be there I am surprised it didn't break anything.

src/command_buffer.rs
```diff
@@ -121,18 +121,13 @@ impl CommandBuffer {
     }
 
     fn build(&mut self, mark: usize) -> (Entity, ReadyBuffer<'_>) {
-        // Amount of components entity has
-        let len = self.entities[mark].end - self.entities[mark].begin;
-        self.ids
-            .extend(self.info[0..len].iter().map(|x| x.ty_info.id()));
+        self.ids.extend(
+            self.info[self.entities[mark].begin..self.entities[mark].end]
```
ouz-a, line 125: I think it's clear now and the comment is no longer needed

src/command_buffer.rs
```diff
@@ -0,0 +1,220 @@
...
+
+/// The output of an '[CommandBuffer]` suitable for passing to
+/// [`World::spawn_into`](crate::World::spawn_into)
+struct ReadyBuffer<'a> {
+    buffer: &'a mut CommandBuffer,
+    mark: usize,
+}
+unsafe impl DynamicBundle for ReadyBuffer<'_> {
+    fn with_ids<T>(&self, f: impl FnOnce(&[TypeId]) -> T) -> T {
+        f(&self.buffer.ids)
```
Ralith, line 185: This passes the *entire* list of IDs, which may have been `extend`ed multiple times. Is that intended?

Given the issues here, I think this module should have a unit test that verifies correct behavior for these methods.

ouz-a: good catch this caused a very sneaky bug, I fixed it and added a test for it, for other methods tests I provided before should do the work.

src/command_buffer.rs
```diff
@@ -121,6 +121,7 @@ impl CommandBuffer {
     }
 
     fn build(&mut self, mark: usize) -> (Entity, ReadyBuffer<'_>) {
+        self.ids.clear();
```
ouz-a, line 124: previously this resulted in generating archetypes for per entity even if they had the same component types,

.gitignore
```diff
@@ -1,3 +0,0 @@
-/target
-**/*.rs.bk
-Cargo.lock
```
Ralith, line 3: Unintended change?

src/command_buffer.rs
```diff
@@ -218,3 +219,25 @@ struct EntityIndex {
...
+        let enta = world.reserve_entity();
+        let entb = world.reserve_entity();
+        let entc = world.reserve_entity();
+        buffer.spawn_at(ent, (true, "a"));
+        buffer.spawn_at(entc, (true, "a"));
+        buffer.spawn_at(enta, (1, 1.0));
+        buffer.spawn_at(entb, (1.0, "a"));
+        buffer.run_on(&mut world);
+        let archetypes = world.archetypes_inner().iter();
+        assert_eq!(archetypes.len(), 4);
```
Ralith, line 241: Shouldn't `ent` and `entc` land in the same archetype, making this 3 (counting the empty archetype)?

Ralith (changes requested):

.gitignore
ouz-a: I don't know how this sneak into here...

src/command_buffer.rs
ouz-a: before anything spawns  world.archetypes_inner().iter().len() gives 1 after 3 spawns it gives 4, I tested this with normal spawns too which gave the same result

Ralith: You're correct; counting is hard.

.gitignore
```diff
@@ -1,3 +1,3 @@
 /target
 **/*.rs.bk
-Cargo.lock
+Cargo.lock
```
Ralith, line 3: This wasn't quite reverted properly--now there's a missing trailing newline.

Ralith (approved):
Thanks!

PR #205, opened 2021-11-16
Replace usage of AtomicU64 and AtomicI64 for MIPS and PPC compatibility

### LINKED ISSUE #204: AtomicU64/AtomicI64 not supported on 32-bit mips/ppc
https://doc.rust-lang.org/std/sync/atomic/#portability

```
PowerPC and MIPS platforms with 32-bit pointers do not have `AtomicU64` or
//!   `AtomicI64` types.
```

You're using AtomicU64/AtomicI64 in two places https://github.com/Ralith/hecs/blob/270e3f60dd2ef2291a774b25c0e34106de7302f5/src/entities.rs#L174
https://github.com/Ralith/hecs/blob/2c1dbcb6d60f7ed83e58677c53ea829cd2780ecb/src/world.rs#L57

Which breaks compatibility for these platforms. 

I worked around it by changing to Atomic[I|U]Size in this patch https://bpa.st/OGYA

If you have another idea to solve, let me know.

sajattack:
fixes #204 

Is this what you had in mind @Ralith ? Not too familiar with the spin mutex you recommended or the initialization your variable needs but I gave it my best shot.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ hecs-macros = { path = "macros", version = "0.7.0", optional = true }
 hashbrown = { version = "0.11.0", default-features = false, features = ["ahash", "inline-more"] }
 lazy_static = { version = "1.4.0", optional = true, features = ["spin_no_std"] }
 serde = { version = "1.0.117", default-features = false, optional = true }
+spin = { version = "0.9.2", default-features=false, features = ["mutex", "spin_mutex"] }
 
 [dev-dependencies]
 bencher = "0.1.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ hecs-macros = { path = "macros", version = "0.7.0", optional = true }
 hashbrown = { version = "0.11.0", default-features = false, features = ["ahash", "inline-more"] }
 lazy_static = { version = "1.4.0", optional = true, features = ["spin_no_std"] }
 serde = { version = "1.0.117", default-features = false, optional = true }
-spin = { version = "0.9.2", default-features=false, features = ["mutex", "spin_mutex"] }
+spin = { version = "0.9.2", default-features = false, features = ["mutex", "spin_mutex"] }
 
 [dev-dependencies]
 bencher = "0.1.5"
--- a/src/world.rs
+++ b/src/world.rs
@@ -58,17 +58,17 @@ impl World {
         // AtomicU64 is unsupported on 32-bit MIPS and PPC architectures
         // For compatibility, use Mutex<u64>
         static ID: Mutex<u64> = Mutex::new(1);
-        {
-            let mut lock_result = ID.lock();
-            let id = lock_result.checked_add(1).unwrap();
-            *lock_result = id;
-
-            Self {
-                entities: Entities::default(),
-                archetypes: ArchetypeSet::new(),
-                bundle_to_archetype: HashMap::default(),
-                id,
-            }
+        let id = {
+            let mut id = ID.lock();
+            let next = id.checked_add(1).unwrap();
+            *id = next;
+            next
+        };
+        Self {
+            entities: Entities::default(),
+            archetypes: ArchetypeSet::new(),
+            bundle_to_archetype: HashMap::default(),
+            id
         }
     }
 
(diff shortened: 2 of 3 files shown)


src/world.rs
```diff
@@ -54,15 +55,13 @@ pub struct World {
...
+        static ID: Mutex<u64> = Mutex::new(1);
 
         Self {
             entities: Entities::default(),
             archetypes: ArchetypeSet::new(),
             bundle_to_archetype: HashMap::default(),
-            id: ID
-                .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |id| id.checked_add(1))
-                .unwrap(),
+            id: *ID.lock() + 1,
```
Ralith, line 64: This just adds 1 to the current value, but leaves that value unchanged, so every invocation will get the same result. We should compute a new value using `checked_add`, panic if it overflows, store the new value, then yield it for use as `id`.

Cargo.toml
```diff
@@ -32,6 +32,7 @@ hecs-macros = { path = "macros", version = "0.7.0", optional = true }
 hashbrown = { version = "0.11.0", default-features = false, features = ["ahash", "inline-more"] }
 lazy_static = { version = "1.4.0", optional = true, features = ["spin_no_std"] }
 serde = { version = "1.0.117", default-features = false, optional = true }
+spin = { version = "0.9.2", default-features=false, features = ["mutex", "spin_mutex"] }
```
Ralith, line 35: Style nit:
```suggestion
spin = { version = "0.9.2", default-features = false, features = ["mutex", "spin_mutex"] }
```

src/world.rs
```diff
@@ -54,15 +55,13 @@ pub struct World {
 impl World {
     /// Create an empty world
     pub fn new() -> Self {
-        static ID: AtomicU64 = AtomicU64::new(1);
+        static ID: Mutex<u64> = Mutex::new(1);
```
Ralith, line 58: Let's add a comment here explaining the motivation (i.e. 32-bit ppc/mips support) for future reference.

Ralith (changes requested):
Thanks!

src/world.rs
```diff
@@ -55,13 +55,18 @@ pub struct World {
 impl World {
     /// Create an empty world
     pub fn new() -> Self {
+        // AtomicU64 is unsupported on 32-bit MIPS and PPC architectures
+        // For compatibility, use Mutex<u64>
         static ID: Mutex<u64> = Mutex::new(1);
 
+        let new_id = ID.lock().checked_add(1).unwrap();
+        *ID.lock() = new_id;
```
Ralith, line 63: `lock`ing twice introduces a race condition; if two worlds are constructed concurrently, both might read the same value from `ID`, increment it, and store the same value back, resulting in `ID` being incremented once and two worlds having the same `id` member. We must lock it once, hold that lock until the updated value is written, and then release it before performing any other work.

Ralith (changes requested):

src/world.rs
sajattack: Sorry, I don't understand how to do that. The only way to get access to the inner u64 that I see is to lock the mutex. So I don't know how to write to it without locking it again.

Ralith: `lock` returns a (smart) reference. Store that reference, then operate on it as many times as you like, then drop it when you're done.

sajattack: Oops, I thought I tried this, but forgot to writeback to ID.

src/world.rs
```diff
@@ -58,15 +58,17 @@ impl World {
...
+            let id = lock_result.checked_add(1).unwrap();
+            *lock_result = id;
+
+            Self {
+                entities: Entities::default(),
+                archetypes: ArchetypeSet::new(),
+                bundle_to_archetype: HashMap::default(),
+                id,
+            }
         }
```
Ralith, line 72: Because this is at the end of a function, wrapping `{}` around it doesn't have any effect. We're also still holding the lock for the duration of the assorted constructors below, when we've already done everything we need. Ideally the lock should be scoped as narrowly as possible, e.g.

```rust
let id = {
    let mut id = ID.lock();
    let next = id.checked_add(1).unwrap();
    *id = next;
    next
};
```

Ralith (approved):
Thanks!

Ralith:
The failure on beta is irrelevant, but it needs a `cargo fmt`.

Issue #177, opened 2021-07-31
Optimize random access


Ralith:
Fetching individual components currently requires hitting the hash table that maps component types to columns every time. For applications which do large amounts of random access for the same component types (e.g. rapier) this is redundant. We could optimize this case by precomputing an archetype -> column offset mapping for a given type up front, in a helper that borrows the `World` and hence will be disposed before any spawns might invalidate it. This would also allow us to perform dynamic borrow checking only once.

Implementation outline:
```rust
impl World {
    fn column<T: Component>(&self) -> Column<'_, T>;
    fn column_mut<T: Component>(&self) -> ColumnMut<'_, T>;
}

struct Column<'a, T> {
    entities: &'a [EntityMeta],
    archetypes: &'a [Archetype],
    archetype_column_offsets: Vec<Option<u32>>,
    _marker: PhantomData<T>,
}

impl<T> Column<'_, T> {
    fn get(&self, entity: Entity) -> Result<&T, ComponentError>;
}
```

Ralith:
It might also be worth cacheing the column offset tables, rather than computing them on-demand. A `TypeIdMap<Vec<Option<u32>>>`, updated whenever a new archetype is added or an existing archetype is realloced, would do the trick.

Ralith:
Implemented in #206.

PR #206, opened 2021-11-19
Optimization for random component access

ouz-a:
This is an optimal solution for issue #177  in my benchmarks I got 30% better performance compared to the native method of getting single component from an entity many times.


benches/bench.rs
```diff
@@ -207,6 +207,21 @@ fn build_cloneable(b: &mut Bencher) {
...
+    let mut world = World::new();
+    let _enta = world.spawn((Position(0.0), Velocity(0.0)));
+    let _entb = world.spawn((true, 12));
+    let entc = world.spawn((Position(3.0),));
+    let _entd = world.spawn((13, true, 4.0));
+    let column = world.column::<Position>();
+
+    b.iter(|| {
+        for i in 0..100 {
+            let comp = column.get(entc).unwrap();
```
Ralith, line 220: We need to ensure this isn't getting optimized out entirely. `bencher::black_box` should hopefully do the trick.

src/column.rs
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    /// Get this column's component from entity if it exist
+    pub fn get(&self, entity: Entity) -> Result<&T, ComponentError> {
+        unsafe {
+            let meta = self
+                .entities
+                .get(entity.id as usize)
+                .expect("no such entity");
```
Ralith, line 38: We should return `ComponentError::NoSuchEntity` rather than panicking here. Additionally, we need to check the `generation` of `entity`. Finally, please narrow the scope of `unsafe` as much as possible; e.g. `[T]::get` is perfectly safe.

src/column.rs
```diff
@@ -0,0 +1,105 @@
...
+use alloc::vec::Vec;
+
+use crate::entities::EntityMeta;
+use crate::{Archetype, Component, ComponentError, Entity, MissingComponent};
+
+/// Column of a single component
+pub struct Column<'a, T> {
+    entities: &'a [EntityMeta],
+    archetypes: &'a [Archetype],
+    archetype_column_offsets: Vec<Option<usize>>,
```
Ralith, line 13: nit:
```suggestion
    archetype_column_indices: Vec<Option<usize>>,
```

src/column.rs
```diff
@@ -0,0 +1,105 @@
...
+                .get(meta.location.archetype as usize)
+                .unwrap();
+            let state = self.archetype_column_offsets[meta.location.archetype as usize]
+                .ok_or_else(MissingComponent::new::<T>)?;
+            let target = NonNull::new_unchecked(
+                archetype
+                    .get_base::<T>(state)
+                    .as_ptr()
+                    .add(meta.location.index as usize),
+            );
```
Ralith, line 50: `NonNull` is only useful when we're storing it somewhere that might allow a niche to be exposed. Just constructing and immediately destructuring it with `as_ptr` doesn't accomplish anything.

src/column.rs
```diff
@@ -0,0 +1,105 @@
...
+            archetype.borrow::<T>(state);
```
Ralith, line 51: This borrow is never released. Additionally, the borrow should be taken inside `new` so we can do it once rather than repeating it for each access.

src/world.rs
```diff
@@ -774,6 +775,41 @@ impl World {
...
+
+        Column::new(entities, archetypes, archetype_column_offsets, PhantomData)
+    }
+
+    /// Faster way of uniquely accessing a Component type
+    ///
+    /// It's useful for applications that do large amount of random
+    /// access to same Component type
+    ///
+    /// #Example
```
Ralith, line 795:
```suggestion
    /// # Example
```

src/world.rs
```diff
@@ -774,6 +775,41 @@ impl World {
...
+    /// ```
+    /// use::hecs::*;
+    /// let mut world = World::new();
+    /// let ent = world.spawn((123,"abc"));
+    /// let column_mut = world.column_mut::<i32>();
+    /// assert_eq!(*column_mut.get(ent).unwrap(),123);
```
Ralith, line 801: It's important to use good style in examples to maximize readability and teach good habits.
```suggestion
    /// let ent = world.spawn((123, "abc"));
    /// let column_mut = world.column_mut::<i32>();
    /// assert_eq!(*column_mut.get(ent).unwrap(), 123);
```

src/world.rs
```diff
@@ -774,6 +775,41 @@ impl World {
         self.archetypes_inner().iter()
     }
 
+    /// Faster way of accessing a Component many times
```
Ralith, line 778: This is *not* a faster way to access a single component. We should convey that it optimizes cases where you're performing repeated random access within a single component type, e.g.:

```suggestion
    /// Efficient random access within a single component type
```

We should also document the conditions under which a dynamic borrowck panic might arise.

src/world.rs
```diff
@@ -774,6 +775,41 @@ impl World {
...
+    pub fn column<T: Component>(&self) -> Column<'_, T> {
+        let archetypes = self.archetypes.archetypes.as_slice();
+        let entities = self.entities.meta.as_slice();
+        let mut archetype_column_offsets: Vec<Option<usize>> = Vec::new();
+        for i in archetypes.iter() {
+            archetype_column_offsets.push(i.get_state::<T>());
+        }
```
Ralith, line 785: Could this work be moved inside `Column::new`? That would help limit the growth of this already large module, and save an argument.

tests/tests.rs
```diff
@@ -584,3 +584,25 @@ fn query_or() {
...
+
+#[test]
+fn column_get_mut() {
+    let mut world = World::new();
+
+    let ent = world.spawn((0, true));
+    let column = world.column_mut::<i32>();
+    let val = column.get(ent).unwrap();
+    *val = 99;
+    assert_eq!(*val, 99);
```
Ralith, line 607: We should use `world.get` to verify that the write affected the correct location through a different path.

benches/bench.rs
```diff
@@ -207,6 +207,21 @@ fn build_cloneable(b: &mut Bencher) {
...
+fn access_column(b: &mut Bencher) {
+    let mut world = World::new();
+    let _enta = world.spawn((Position(0.0), Velocity(0.0)));
+    let _entb = world.spawn((true, 12));
+    let entc = world.spawn((Position(3.0),));
+    let _entd = world.spawn((13, true, 4.0));
+    let column = world.column::<Position>();
+
+    b.iter(|| {
+        for i in 0..100 {
```
Ralith, line 219: Why loop here?

Ralith (changes requested):

qthree:
Since `Column` is already `!Send` you can use `Cell` and fill `archetype_column_offsets` lazily upon first get access to concrete `Archetype`.
```rust
archetype_column_offsets: Vec<Cell<Poll<Option<usize>>>>,
```
This way you'll also know which archetypes you've already borrowed with `archetype.borrow::<T>(state);`.

benches/bench.rs
ouz-a: > For applications which do large amounts of random access for the same component types
the issue stated this as a problem, so I wanted to test _large access_ 😸

benches/bench.rs
```diff
@@ -217,7 +217,7 @@ fn access_column(b: &mut Bencher) {
 
     b.iter(|| {
         for i in 0..100 {
-            let comp = column.get(entc).unwrap();
+            let comp = bencher::black_box(column.get(entc).unwrap());
```
ouz-a, line 220: I don't even know if this works now ? or how to test it

src/column.rs
```diff
@@ -1,105 +1,152 @@
...
 
 impl<'a, T: Component> ColumnMut<'a, T> {
     pub(crate) fn new(
         entities: &'a [EntityMeta],
         archetypes: &'a [Archetype],
-        archetype_column_offsets: Vec<Option<usize>>,
         _marker: PhantomData<T>,
     ) -> Self {
+        let mut archetype_column_indices: Vec<Option<usize>> = Vec::new();
+        for i in archetypes.iter() {
```
ouz-a, line 97: this change made Column quite faster but now it borrows every component T, before it only borrowed from a given entity. I am not sure if this is a good thing or bad thing

src/column.rs
```diff
@@ -1,105 +1,152 @@
...
             _marker,
         }
     }
 
     /// Mutably get this column's component from entity if it exist
     pub fn get(&self, entity: Entity) -> Result<&mut T, ComponentError> {
+        let meta = self
+            .entities
+            .get(entity.id as usize)
+            .map_or(true, |meta| meta.generation == entity.generation)
```
ouz-a, line 119: very verbose but checks the generation

benches/bench.rs
Ralith: `iter` will already run the supplied closure many times to compute the average performance, so this isn't necessary.

benches/bench.rs
Ralith: Only way to know for sure is to look at the output assembly or IR, but at least we're making an effort now. Please also resolve the lint.

src/column.rs
Ralith: The purpose of this interface is specifically to optimize for very heavy random access, so I think this is the correct tradeoff.

src/column.rs
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+
+    /// Mutably get this column's component from entity if it exist
+    pub fn get(&self, entity: Entity) -> Result<&mut T, ComponentError> {
+        let meta = self
+            .entities
+            .get(entity.id as usize)
+            .map_or(true, |meta| meta.generation == entity.generation)
+            .then(|| self.entities[entity.id as usize])
```
Ralith, line 120: This will panic if `entity.id` is out of range; it should instead fail with `NoSuchEntity`.

Ralith (changes requested):

src/column.rs
```diff
@@ -0,0 +1,152 @@
...
+            .get(meta.location.archetype as usize)
+            .unwrap();
+        let state = self.archetype_column_indices[meta.location.archetype as usize]
+            .ok_or_else(MissingComponent::new::<T>)?;
+        unsafe {
+            let target = archetype
+                .get_base::<T>(state)
+                .as_ptr()
+                .add(meta.location.index as usize)
+                .cast::<T>();
```
Ralith, line 60: This `cast` invocation shouldn't be necessary.

src/column.rs
```diff
@@ -44,8 +44,8 @@ impl<'a, T: Component> Column<'a, T> {
             .entities
             .get(entity.id as usize)
             .map_or(true, |meta| meta.generation == entity.generation)
-            .then(|| self.entities[entity.id as usize])
-            .ok_or(NoSuchEntity)?;
+            .then(|| self.entities.get(entity.id as usize).ok_or(NoSuchEntity))
+            .ok_or(NoSuchEntity)??;
```
ouz-a, line 48: I am really bad at functional style 😭

ouz-a:
Screen Shot 2021-11-23 at 00 00 07

I wrote a basic function to compare two access types and columnar access as expected is significantly faster

src/column.rs
Ralith: Sorry, I skimmed this a little too quickly. The real (mostly cosmetic) issue here is that we're indexing into `entities` twice. A tidier approach would be:
```rust
let meta = self
    .entities
    .get(entity.id as usize)
    .filter(|meta| meta.generation == entity.generation)
    .ok_or(NoSuchEntity)?;
```

Ralith (changes requested):

Ralith (approved):
Thanks!

PR #207, opened 2021-11-24
Make TypeInfo's accessors public and add a from_parts constructor

sdleffler:
Does what it says on the tin.

### DIFF
--- a/src/archetype.rs
+++ b/src/archetype.rs
@@ -494,10 +494,11 @@ impl<V> OrderedTypeIdMap<V> {
     }
 }
 
-/// Metadata required to store a component. All told, this means a [`TypeId`], to be able to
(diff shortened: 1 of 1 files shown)


src/archetype.rs
```diff
@@ -494,7 +494,10 @@ impl<V> OrderedTypeIdMap<V> {
     }
 }
 
-/// Metadata required to store a component
+/// Metadata required to store a component. All told, this means a [`TypeId`], to be able to
+/// dynamically name/check the component type; a [`Layout`], so that we know how to allocate memory
```
Ralith, line 498: There should be an empty line here after the first sentence.

sdleffler: Added.

Ralith (approved):
Thanks!

PR #209, opened 2021-11-26
Leave World in a consistent state after query borrowck panics

Ralith:
Previously, some archetypes might be left permanently borrowed, making the World unusable.


Ralith:
TODO: The current draft unwinds borrows of archetypes which the query has already fully borrowed from. We also need a way to unwind the incomplete borrow of the archetype on which borrow checking failed. That'll probably have to be handled in every aggregating `Fetch::borrow` implementation.

adamreichold:
> We also need a way to unwind the incomplete borrow of the archetype on which borrow checking failed

I think the cleanest way of doing this would be drop guards like the `Ref(Mut)` types exposed by `RefCell`. But this would also imply having a dynamic allocation around to store those in which is probably alright for `PreparedQuery` but may be too expensive for ad-hoc `Query`.

adamreichold:
@Ralith I think there is another option for using `Drop` guards but without the requirement for dynamic allocation: Using global instead of per-archetype borrow flags, c.f. https://github.com/adamreichold/rs-ecs/pull/28.

This does reduce borrow checking overhead for queries, but it does also change semantics, e.g.
```rust
let foo_bar = world.get_mut::<Foo>(bar);
let foo_qux = world.get_mut::<Foo>(qux);
```
would work now if it is guaranteed that `bar` and `qux` are in distinct archetypes. for example by being distinct kinds of entities which can never have fully overlapping components. On the other hand, archetypes could be viewed as an implementation detail so that user code should not rely on constructs like the above and use e.g. `View::get_mut_n`.

Issue #211, opened 2021-11-28
Improvement: Add BundleClone and EntityBuilderClone::add_bundle


ten3roberts:
Currently, there is no add_bundle or alike function to the cloneable entity builder due to the fact that there is no trait for a cloneable bundle of components or similar.

A helper trait similar to Bundle which is auto implemented for tuples that implement clone would enable this behaviour and could be published without breakages.

Ralith:
Released in 0.7.1.

PR #212, opened 2021-11-28
Add EntityBuilderClone::add_bundle and tests

ten3roberts:
This fixes issue #211 without any apparent breakages in the public API, allowing for an immediate patch release.

The commit also adds appropriate tests for the cloneable entity builder which were previously missing.

### DIFF
--- a/src/bundle.rs
+++ b/src/bundle.rs
@@ -6,6 +6,7 @@
 // copied, modified, or distributed except according to those terms.
 
 use crate::alloc::{vec, vec::Vec};
+use crate::entity_builder::Cloneable;
 use core::any::{type_name, TypeId};
 use core::ptr::NonNull;
 use core::{fmt, mem};
@@ -57,6 +58,17 @@ pub unsafe trait Bundle: DynamicBundle {
         Self: Sized;
 }
 
+/// A dynamically typed collection of components that is also cloneable
+pub unsafe trait DynamicBundleClone: DynamicBundle + Clone {
+    /// Allow a callback to move all components out of the bundle
+    ///
+    /// Must invoke `f` only with a valid pointer and the pointee's type and size. `put` may only be
+    /// called at most once on any given value. Provides helper type for cloning
+    /// each type.
+    #[doc(hidden)]
+    unsafe fn put_with_clone(self, f: impl FnMut(*mut u8, TypeInfo, Cloneable));
+}
+
 /// Error indicating that an entity did not have a required component
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct MissingComponent(&'static str);
@@ -100,6 +112,23 @@ macro_rules! tuple_impl {
                     f(
                         (&mut $name as *mut $name).cast::<u8>(),
                         TypeInfo::of::<$name>()
+                    );
+                    mem::forget($name);
+                )*
+            }
+        }
+
+        unsafe impl<$($name: Component + Clone),*> DynamicBundleClone for ($($name,)*) {
+            // Compiler false positive warnings
+            #[allow(unused_variables, unused_mut)]
+            unsafe fn put_with_clone(self, mut f: impl FnMut(*mut u8, TypeInfo, Cloneable)) {
+                #[allow(non_snake_case)]
+                let ($(mut $name,)*) = self;
+                $(
+                    f(
+                        (&mut $name as *mut $name).cast::<u8>(),
+                        TypeInfo::of::<$name>(),
+                        Cloneable::new::<$name>(TypeInfo::of::<$name>())
                     );
                     mem::forget($name);
                 )*
--- a/src/bundle.rs
+++ b/src/bundle.rs
@@ -128,7 +128,7 @@ macro_rules! tuple_impl {
                     f(
                         (&mut $name as *mut $name).cast::<u8>(),
                         TypeInfo::of::<$name>(),
-                        Cloneable::new::<$name>(TypeInfo::of::<$name>())
+                        Cloneable::new::<$name>()
                     );
                     mem::forget($name);
                 )*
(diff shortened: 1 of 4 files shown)


src/entity_builder.rs
```diff
@@ -374,15 +387,24 @@ impl<M> Default for Common<M> {
...
-        Self(|src, f| unsafe {
-            let mut tmp = (*src.cast::<T>()).clone();
-            f((&mut tmp as *mut T).cast(), TypeInfo::of::<T>());
-            core::mem::forget(tmp);
-        })
+    pub(crate) fn new<T: Component + Clone>(type_info: TypeInfo) -> Self {
+        Self {
+            type_info,
+            func: |src, ty, f| unsafe {
+                std::dbg!(ty);
```
Ralith, line 402: Stray `dbg`

src/world.rs
```diff
@@ -159,6 +159,7 @@ impl World {
             components.put(|ptr, ty| {
                 archetype.put_dynamic(ptr, ty.id(), ty.layout().size(), index);
             });
+
```
Ralith, line 162: Stray edit

tests/tests.rs
```diff
@@ -200,6 +200,31 @@ fn build_entity() {
     assert_eq!(*world.get::<i32>(f).unwrap(), 789);
 }
 
+#[test]
+fn build_entity_clone() {
+    let mut world = World::new();
+    let mut entity = EntityBuilderClone::new();
+    entity.add("abc");
+    entity.add(123);
+    let e = world.spawn(&entity.build());
```
Ralith, line 209: What's the test coverage benefit of constructing two different builders?

tests/tests.rs
```diff
@@ -200,6 +200,31 @@ fn build_entity() {
...
+    let mut entity = EntityBuilderClone::new();
+    entity.add("def");
+    entity.add([0u8; 1024]);
+    entity.add(456);
+    entity.add(789);
+    entity.add_bundle(("yup", 67_usize));
+    entity.add_bundle((5.0_f32, String::from("Foo")));
+    entity.add_bundle((7.0_f32, String::from("Bar"), 42_usize));
+    let f = world.spawn(&entity.build());
```
Ralith, line 218: Let's spawn multiple times so that any inadvertent memcpying is more likely to trigger a crash.

src/entity_builder.rs
```diff
@@ -374,15 +387,24 @@ impl<M> Default for Common<M> {
 }
 
 #[derive(Clone)]
-struct Cloneable(unsafe fn(*const u8, &mut dyn FnMut(*mut u8, TypeInfo)));
+/// Struct providing dynamic clone via callback
+pub struct Cloneable {
+    func: unsafe fn(*const u8, TypeInfo, &mut dyn FnMut(*mut u8, TypeInfo)),
+    // Store type info along with cloneable
+    type_info: TypeInfo,
```
Ralith, line 394: Why the change to store this rather than constructing it on-demand?

Ralith (changes requested):
Thanks! I agree this is a very natural extension.

src/entity_builder.rs
ten3roberts: You are right. If typeinfo is cheap to construct it is not necessary to store it.

ten3roberts: It was left over from storing the type info in an earlier attempt. Since it is known with bundle::put I thought to store it rather than compute again.

ten3roberts (commented):
Great at spotting these.

tests/tests.rs
ten3roberts: This follows the style of `built_entity` test. It may increase the coverage as one with many modifications and one with few modifications of already existing components. Will add testing for spawning multiple entities from the same builder

src/entity_builder.rs
ten3roberts: The function is from my understanding called for each build

Ralith: `TypeInfo::of` will almost certainly be compiled out into a constant, so storing `TypeInfo`s will probably *add* overhead, in addition to complicating the code. As usual, beware premature optimizations.

src/bundle.rs
```diff
@@ -57,6 +58,17 @@ pub unsafe trait Bundle: DynamicBundle {
         Self: Sized;
 }
 
+/// A dynamically typed collection of components that is also cloneable
+pub unsafe trait DynamicBundleClone: DynamicBundle + Clone {
```
Ralith, line 62: I think this is slightly more strict than necessary; we don't actually need to be able to clone the bundle itself, we just need to know how to clone its contents.
```suggestion
/// A dynamically typed collection of cloneable components
pub unsafe trait DynamicBundleClone: DynamicBundle {
```

Ralith (changes requested):

src/entity_builder.rs
```diff
@@ -173,6 +174,18 @@ impl EntityBuilderClone {
         self
     }
 
+    /// Add all components in `bundle` to the entity.
+    ///
+    /// If the bundle contains any component which matches the type of a component
+    /// already in the `EntityBuilder`, the newly added component from the bundle
+    /// will replace the old component and the old component will be dropped.
+    pub fn add_bundle<T: Component + DynamicBundleClone>(&mut self, bundle: T) -> &mut Self {
```
Ralith, line 182: `Component` is overly strict here, and let's use `impl Trait` for consistency with `EntityBuilder`.
```suggestion
    pub fn add_bundle(&mut self, bundle: impl DynamicBundleClone) -> &mut Self {
```

Ralith (changes requested):

src/bundle.rs
ten3roberts: That is correct. Can't think of an implementation where those aren't transitive properties unless a custom implementation.

ten3roberts:
All tests pass after introducing the changes

Ralith (approved):
Thanks!

Issue #214, opened 2021-11-29
[ macro ] Make hecs-macros::Bundle implement DynamicBundleClone


ten3roberts:
If a struct implements `Clone` the `Bundle` procedural macro should change behaviour and also implement `DynamicBundleClone`, allowing for seamless usage of bundle structs in cloneable builders. 

This change would not affect `hecs`, only `hecs-macros`

ten3roberts:
In my game engine `ivy`, I have multiple bundlez such as `TransformBundle`, or `RbBundle` that allows initialization of required components with sane defaults, such as adding a scale of 1. This allows me to not having to use Option queries and instead rely on sane defaults if not all fields are specified. It also allows the components to be mutated because they exist. An example of this is a Resitution value for colliders which is often left at default, but could be change without having to check and add it if it doesn't. It exists for all entities also having a velocity etc.

These bundles are all value types with Copy, Clone, PartialEq etc, but as of currently cannot be used with `EntityBuilderClone`.

I do not have much experience with procedural macros, so I do not think I may be able to implement it myself.

Ralith:
A macro cannot judge whether a struct('s field) implements `Clone`, because they run before type-checking. Instead, a new `derive(DynamicBundleClone)` macro should be defined.

ten3roberts:
In that case I can start looking into it. Should `DynamicBundleClone` only implement only DynamicBundleClone or Bundle and DynamicBundle as well?

In my opinion, the first option is better and requiring both `Bundle` and `DynamicBundleClone` derives which ensures implementations won't conflict if both are used. This is similar to the std PartialEq, Eq, and Eq, Hash pairs.

Ralith:
Fixed by #215.

PR #215, opened 2021-11-29
Add `DynamicBundleClone` derive macro

ten3roberts:
Note: `DynamicClone::new` has to be visibile outside the crate in order to be used in the macro.

Fixes issue #214

### DIFF
--- a/macros/src/bundle_clone.rs
+++ b/macros/src/bundle_clone.rs
@@ -1,7 +1,8 @@
-use proc_macro2::Span;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
 use syn::{DeriveInput, Error, Result};
+
+use crate::common::struct_fields;
 
 pub fn derive(input: DeriveInput) -> Result<TokenStream2> {
     let ident = input.ident;
@@ -60,29 +61,4 @@ fn add_additional_bounds_to_generic_params(mut generics: syn::Generics) -> syn::
             .push(syn::TypeParamBound::Trait(make_component_trait_bound()))
     });
     generics
-}
-
-fn struct_fields(fields: &syn::Fields) -> (Vec<&syn::Type>, Vec<syn::Member>) {
-    match fields {
-        syn::Fields::Named(ref fields) => fields
-            .named
-            .iter()
-            .map(|f| (&f.ty, syn::Member::Named(f.ident.clone().unwrap())))
-            .unzip(),
-        syn::Fields::Unnamed(ref fields) => fields
-            .unnamed
-            .iter()
-            .enumerate()
-            .map(|(i, f)| {
-                (
-                    &f.ty,
-                    syn::Member::Unnamed(syn::Index {
-                        index: i as u32,
-                        span: Span::call_site(),
-                    }),
-                )
-            })
-            .unzip(),
-        syn::Fields::Unit => (Vec::new(), Vec::new()),
-    }
 }
--- a/macros/src/bundle_clone.rs
+++ b/macros/src/bundle_clone.rs
@@ -1,0 +1,88 @@
+use proc_macro2::Span;
+use proc_macro2::TokenStream as TokenStream2;
+use quote::quote;
+use syn::{DeriveInput, Error, Result};
+
+pub fn derive(input: DeriveInput) -> Result<TokenStream2> {
+    let ident = input.ident;
+    let data = match input.data {
+        syn::Data::Struct(s) => s,
+        _ => {
+            return Err(Error::new_spanned(
+                ident,
+                "derive(DynamicBundleClone) does not support enums or unions",
+            ))
+        }
+    };
+    let (tys, field_members) = struct_fields(&data.fields);
+    let generics = add_additional_bounds_to_generic_params(input.generics);
+
+    let dyn_bundle_code = gen_dynamic_bundle_impl(&ident, &generics, &field_members, &tys);
+    Ok(dyn_bundle_code)
+}
+
+fn gen_dynamic_bundle_impl(
+    ident: &syn::Ident,
+    generics: &syn::Generics,
+    field_members: &[syn::Member],
+    tys: &[&syn::Type],
+) -> TokenStream2 {
+    let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
+    quote! {
+        unsafe impl #impl_generics ::hecs::DynamicBundleClone for #ident #ty_generics #where_clause {
+            #[allow(clippy::forget_copy)]
+            unsafe fn put_with_clone(mut self, mut f: impl ::std::ops::FnMut(*mut u8, ::hecs::TypeInfo, DynamicClone)) {
+                #(
+                    f(
+                        (&mut self.#field_members as *mut #tys).cast::<u8>(),
+                        ::hecs::TypeInfo::of::<#tys>(),
+                        ::hecs::DynamicClone::new::<#tys>()
+                    );
+                    ::std::mem::forget(self.#field_members);
+                )*
+            }
+        }
+    }
+}
+
+fn make_component_trait_bound() -> syn::TraitBound {
+    syn::TraitBound {
+        paren_token: None,
+        modifier: syn::TraitBoundModifier::None,
+        lifetimes: None,
+        path: syn::parse_quote!(::hecs::ComponentClone),
+    }
+}
+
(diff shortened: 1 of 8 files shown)


ten3roberts:
I hope I am of help, and if there are any other features that you'd want implemented or have planned I can look into them.

ten3roberts:
There is some code duplication, specifically  `make_structs_fields`. Should I create a crate only common module for reuse?

ten3roberts:
I am wondering if this feature is still wanted, and if my implementation is acceptable to the desired API

Ralith:
Thanks for the PR! I'm interested in this feature, I just haven't had time for open source code review recently.

src/world.rs
```diff
@@ -935,6 +935,14 @@ impl From<NoSuchEntity> for QueryOneError {
 pub trait Component: Send + Sync + 'static {}
 impl<T: Send + Sync + 'static> Component for T {}
 
+/// Types that can be cloneable components, implemented automatically for all
+/// `Send + Sync + Clone + 'static` types
+///
+/// This is just a convenient shorthand for `Send + Sync + Clone + 'static`, and never needs to be
+/// implemented manually.
+pub trait ComponentClone: Send + Sync + Clone + 'static {}
+impl<T: Send + Sync + Clone + 'static> ComponentClone for T {}
```
Ralith, line 944: Does this have any advantages over `Component + Clone`?

macros/src/bundle_clone.rs
```diff
@@ -0,0 +1,88 @@
...
+        lifetimes: None,
+        path: syn::parse_quote!(::hecs::ComponentClone),
+    }
+}
+
+fn add_additional_bounds_to_generic_params(mut generics: syn::Generics) -> syn::Generics {
+    generics.type_params_mut().for_each(|tp| {
+        tp.bounds
+            .push(syn::TypeParamBound::Trait(make_component_trait_bound()))
+    });
```
Ralith, line 62: nit: favor explicit `for` loops over `for_each`.

macros/src/lib.rs
```diff
@@ -40,6 +41,22 @@ pub fn derive_bundle(input: TokenStream) -> TokenStream {
     .into()
 }
 
+/// Implement `DynamicBundleClone` for a struct.
+///
+/// This is an extension macro for bundles which allow them to be cloned, and
+/// subsequently used in `EntityBuilderClone::add_bundle`.
+///
+/// Requires that all fields of the struct implement clone.
```
Ralith, line 51: s/clone/[\`Clone\`]/, and we should mention that `Bundle` must also be derived.

src/lib.rs
```diff
@@ -73,7 +73,7 @@ mod world;
 
 pub use archetype::{Archetype, ArchetypeColumn};
 pub use batch::{BatchIncomplete, BatchWriter, ColumnBatch, ColumnBatchBuilder, ColumnBatchType};
-pub use bundle::{Bundle, DynamicBundle, MissingComponent};
+pub use bundle::{Bundle, DynamicBundle, DynamicBundleClone, DynamicClone, MissingComponent};
```
Ralith, line 76: Since `DynamicClone` is only useful for implementing `#[doc(hidden)]` interfaces ,its reexport should itself be `#[doc(hidden)]`, in the commented block below for that purpose.

Ralith (changes requested):

ten3roberts:
May I ask? Why do you prefer explicit for rather that for_each?

I have never been able to get a good answer on it, and would like for your reasons.

Ralith:
It's a bit clearer because a for loop is always guaranteed to do a specific thing, whereas a method named `for_each` might not. Further, for loops are common in rust code in general, and so by favoring them we produce more conventionally styled code, which is again easier to recognize at a glance.

src/bundle.rs
```diff
@@ -67,6 +67,7 @@ pub unsafe trait DynamicBundleClone: DynamicBundle {
 }
 
 #[derive(Copy, Clone)]
+#[doc(hidden)]
```
Ralith, line 70: This must be applied to the reexport in lib.rs, not the original definition. That reexport should also be moved down into the dedicated block of reexports for the macro crate.

src/world.rs
```diff
@@ -941,7 +941,7 @@ impl<T: Send + Sync + 'static> Component for T {}
 /// This is just a convenient shorthand for `Send + Sync + Clone + 'static`, and never needs to be
 /// implemented manually.
 pub trait ComponentClone: Send + Sync + Clone + 'static {}
-impl<T: Send + Sync + Clone + 'static> ComponentClone for T {}
+impl<T: Component + Clone> ComponentClone for T {}
```
Ralith, line 944: What is the benefit of this trait compared to using `Component + Clone` directly?

macros/src/lib.rs
Ralith: "clone" still needs to be replaced with [\`Clone\`].

Ralith (changes requested):

src/world.rs
ten3roberts: No benifit. Didn't think of using the alias

src/lib.rs
```diff
@@ -99,7 +99,10 @@ pub use lazy_static;
 pub use query::Fetch;
 
 #[cfg(feature = "macros")]
-pub use hecs_macros::{Bundle, Query};
+#[doc(hidden)]
+pub use hecs_macros::DynamicClone;
```
Ralith: `DynamicClone` is defined in this crate, not `hecs_macros`, and the reexport should be part of the block that is commented with `// Unstable implementation details needed by the macros`.

src/world.rs
Ralith: In that case, let's not complicate the public API unnecessarily with an additional trait.

Ralith (changes requested):

src/lib.rs
ten3roberts: Alright

macros/src/lib.rs
```diff
@@ -48,7 +48,7 @@ pub fn derive_bundle(input: TokenStream) -> TokenStream {
 /// This is an extension macro for bundles which allow them to be cloned, and
 /// subsequently used in `EntityBuilderClone::add_bundle`.
 ///
-/// Requires that all fields of the struct implement clone.
+/// Requires that all fields of the struct implement Clone.
```
Ralith, line 51: Because `Clone` is a resolveable identifier, it should be referenced with an intra-doc link, using the syntax [\`Clone\`].

src/lib.rs
```diff
@@ -92,14 +92,17 @@ pub use world::{
...
+pub use bundle::DynamicClone;
 #[cfg(feature = "macros")]
 #[doc(hidden)]
 pub use lazy_static;
 #[doc(hidden)]
 pub use query::Fetch;
 
 #[cfg(feature = "macros")]
-pub use hecs_macros::{Bundle, Query};
+#[cfg(feature = "macros")]
```
Ralith, line 104: Duplicate cfg guard.

macros/src/lib.rs
ten3roberts: Thank you. Learning a lot of Rust here.

macros/src/bundle_clone.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+}
+
+fn make_component_trait_bound() -> syn::TraitBound {
+    syn::TraitBound {
+        paren_token: None,
+        modifier: syn::TraitBoundModifier::None,
+        lifetimes: None,
+        path: syn::parse_quote!(::hecs::ComponentClone),
```
Ralith, line 54: This is now a stale reference, which reveals missing test coverage for this case.

ten3roberts:
I am getting this error when running tests

image

From what I can see it is due to an expansion of a DIR variable, though I do not know what that means

ten3roberts:
Apparently, the test failure is no longer reproducible after restarting my terminal

macros/src/bundle_clone.rs
Ralith: Specifically, we are still missing test coverage for cases like:
```rust
#[derive(DynamicBundleClone, Bundle)]
struct Foo<T> { x: T }
```

Ralith:
Error message matching is unfortunately a bit flakey. Don't pay the trybuild stuff much mind so long as it's passing in CI.

Ralith (approved):
Thanks!

PR #222, opened 2021-12-23
Add query map providing random access to query results

adamreichold:
This is basically a generalisation of the `World::column(_mut)` API for arbitrary queries involving multiple components and/or query combinators.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -11,7 +11,7 @@ use core::mem;
 use core::ptr::NonNull;
 use core::slice::Iter as SliceIter;
 
-use crate::alloc::boxed::Box;
+use crate::alloc::{boxed::Box, vec::Vec};
 use crate::archetype::Archetype;
 use crate::entities::EntityMeta;
 use crate::{Component, Entity, World};
@@ -592,6 +592,12 @@ impl<'w, Q: Query> QueryBorrow<'w, Q> {
         unsafe { QueryIter::new(self.meta, self.archetypes.iter()) }
     }
 
+    /// Provide random access to the query results
+    pub fn view(&mut self) -> View<'_, Q> {
+        self.borrow();
+        unsafe { View::new(self.meta, self.archetypes) }
+    }
+
     /// Like `iter`, but returns child iterators of at most `batch_size` elements
     ///
     /// Useful for distributing work over a threadpool.
@@ -785,6 +791,11 @@ impl<'q, Q: Query> QueryMut<'q, Q> {
         Self {
             iter: unsafe { QueryIter::new(meta, archetypes.iter()) },
         }
+    }
+
+    /// Provide random access to the query results
+    pub fn view(&mut self) -> View<'_, Q> {
+        unsafe { View::new(self.iter.meta, self.iter.archetypes.as_slice()) }
     }
 
     /// Transform the query into one that requires a certain component without borrowing it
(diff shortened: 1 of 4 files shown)


src/query.rs
```diff
@@ -1113,6 +1131,14 @@ impl<'q, Q: Query> PreparedQueryBorrow<'q, Q> {
 
         unsafe { PreparedQueryIter::new(self.meta, self.archetypes, state.iter()) }
     }
+
+    /// Provides random access to the results of the prepared query
+    pub fn map<'i>(&'i mut self) -> PreparedQueryMap<'i, Q> {
```
adamreichold, line 1136: This is currently only defined for `PreparedQuery` as this would allow borrowing the `Vec<Option<Q::Fetch>>` within `PreparedQueryMap` to amortize its allocation costs.

I am not sure what the preferred options for hecs are:
* Should I make the current implementation of `PreparedQueryMap` available as `QueryMap` based off of `Query`?
* Should I put a `Box<[Q::Fetch]>` into `PreparedQuery` to amortize the allocation costs but making `PreparedQuery` larger unconditionally?

src/query.rs
```diff
@@ -1079,6 +1097,23 @@ impl<Q: Query> PreparedQuery<Q> {
 
         unsafe { PreparedQueryIter::new(meta, archetypes, state.iter()) }
     }
+
+    /// Provide random access to query results for a uniquely borrow world
+    pub fn query_map_mut<'q>(&'q mut self, world: &'q mut World) -> PreparedQueryMap<'q, Q> {
```
adamreichold, line 1102: I am not particularly fond of this name and would be glad for a suggestion for a better one.

adamreichold:
Performance is as expected, i.e.
```
> cargo bench --bench bench --all-features -- access
   Compiling hecs v0.7.1 (/home/adam/Projekte/hecs)
    Finished bench [optimized + debuginfo] target(s) in 35.03s
     Running unittests (target/release/deps/bench-4fb23b93b457eec3)

running 2 tests
test access_column                  ... bench:           7 ns/iter (+/- 0)
test access_query_map               ... bench:           1 ns/iter (+/- 0)

test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured
```

I think `Column(Mut)::get` could be faster if it were to store the base pointers instead of the state?

adamreichold:
> I think Column(Mut)::get could be faster if it were to store the base pointers instead of the state?

The last commit brings them on an equal footing:
```
> cargo bench --bench bench --all-features -- access
   Compiling hecs v0.7.1 (/home/adam/Projekte/hecs)
    Finished bench [optimized + debuginfo] target(s) in 34.82s
     Running unittests (target/release/deps/bench-4fb23b93b457eec3)

running 2 tests
test access_column                  ... bench:           1 ns/iter (+/- 0)
test access_query_map               ... bench:           1 ns/iter (+/- 0)

test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured
```

If you'd like I could split that one out into a separate PR as it is independent of the design questions w.r.t. `QueryMap`?

Ralith:
If you make such a PR I'll happily merge it, but if we're going to replace `Column(Mut)` with this anyway I'm not sure it's worth the effort.

src/query.rs
```diff
@@ -65,6 +65,10 @@ pub unsafe trait Fetch<'a>: Sized {
     unsafe fn get(&self, n: usize) -> Self::Item;
 }
 
+/// Marker trait indicating whether a given [`Fetch`] implementation will produce only shared references
```
Ralith, line 68: "will not produce unique references" might be a little more accurate, since e.g. `Satisfies` produces non-reference data.

src/query.rs
Ralith: I don't have any problem with making `PreparedQuery` slightly larger, though amortizing that allocation may be a premature opt since it's off the critical hot path of `get`.

I would like to support non-prepared queries as well for ergonomics. If it's desirable to borrow from `PreparedQuery`, perhaps `Cow` could be used to handle both cases gracefully in a single type?

src/query.rs
```diff
@@ -65,6 +65,10 @@ pub unsafe trait Fetch<'a>: Sized {
...
+#[allow(clippy::missing_safety_doc)]
+pub unsafe trait FetchShared {}
```
Ralith, line 70: Could we mark the query types themselves, rather than the `Fetch` impls? That should be a bit more intuitive for users.

src/query.rs
Ralith: What if we globally replaced "query map" with "view"? It's more concise, and IMO at least as clear.

Ralith (commented):
Thanks; this is much more concise than I expected!

src/query.rs
adamreichold: Changed as suggested.

src/query.rs
adamreichold: Changed.

src/query.rs
adamreichold: > I would like to support non-prepared queries as well for ergonomics. If it's desirable to borrow from PreparedQuery, perhaps Cow could be used to handle both cases gracefully in a single type?

I tried that but it does not really help as it is not just how the `Fetch` instances are stored but also that the states need to be prepared instead of just iterated. I will try adding separate `View` and `PreparedView` types instead.

src/query.rs
adamreichold: Renamed it, it is called `struct View` now and this method became `view_mut`.

src/query.rs
adamreichold: Scratch that, I did make the structure generic over how the fetch instances are stored with two different constructor methods `new` and `new_prepared`.

I went for `AsRef<[Option<Q::Fetch>]>` instead of `Cow<'q, [Option<Q::Fetch>]>` though as the dispatch over the `Cow`'s discriminant would end up in the fast path for `get(_mut)` even though the type of storage is fixed after construction. This does have the downside of exposing the storage in the return type of the public methods though.

Ralith: > the dispatch over the Cow's discriminant would end up in the fast path

Could we cache a raw pointer to the contents?

Ralith: Alternatively, the structure is out small so I think having two almost-duplicates of it isn't the end in the world. In other locations (e.g. `EntityBuilder` vs. `EntityBuilderClone`) I've similarly favored duplication over wonky interfaces.

Ralith:
LGTM aside from the outstanding questions about the view type; thanks!

src/query.rs
adamreichold: I agree. Internal code duplication seems preferable to exposing internals. I modifying the last commit to take that route instead.

Ralith (approved):
Great stuff, thanks!

PR #223, opened 2021-12-23
Optimise Column(Mut

adamreichold:
Use that the base pointer depends only on the archetype but not on the entity within it in the implemenation of Column(Mut).


Ralith:
Thanks!

PR #224, opened 2021-12-24
Add World::exchange as a optimised combination of World::remove followed by World::insert

adamreichold:


adamreichold:
I see at least two open issues here:
* [ ] How much of the copy-and-pasted code from `World::remove` and `World::insert` should be factored out into reusable helper functions?
* [ ] Should there be `exchange_edges` instead of querying `remove_edges` and `insert_edges`?

adamreichold:
The included benchmark improves from
```
test exchange                       ... bench:      29,769 ns/iter (+/- 731)
```
using the "fake" implementation to
```
test exchange                       ... bench:      18,183 ns/iter (+/- 281)
```

Ralith:
Nice speedup!

> How much of the copy-and-pasted code from World::remove and World::insert should be factored out into reusable helper functions?

Given that it's somewhat fraught unsafe stuff, I'd definitely like to minimize duplication here as much as is feasible. It hasn't needed to be touched in a good while, but when it was fresh it was a source of several serious bugs.

> Should there be exchange_edges instead of querying remove_edges and insert_edges?

If there's a meaningful benchmark benefit, I don't see why not.

adamreichold:
> > Should there be exchange_edges instead of querying remove_edges and insert_edges?

> If there's a meaningful benchmark benefit, I don't see why not.

I pushed a commit implementing this for you to have a look at.

I think it is a tough sell: While the benchmark improves from
```
test exchange                       ... bench:      21,497 ns/iter (+/- 487)
```
to
```
test exchange                       ... bench:      19,907 ns/iter (+/- 611)
```
(different baseline to above as I remembered to disable frequency boost now), this requires another specialised hasher `TypeIdPairHasher` (as I cannot just use `TypeId::of::<(S,T)>` due to dynamic bundles), reduces the amount of code that can be shared with `World::{insert,remove}`, increases memory usage due to the extra hash tables and does not implicitly "discover" the intermediate edges for use with `World::{insert,remove}`.

Do you think the 7.5% improvement is worth the additional complexity?

(The additional hasher could probably be avoided by adding something like `Bundle::key_with<T>(&self) -> Option<TypeId>` which basically returns `TypeId::of::<(Self, T)>()`, but that would push the complexity further out into the user-facing API.)

adamreichold:
As a bit of a tangent, I do use separate `{insert,remove,exchange}_map` in rs-ecs but the situation is slightly different because I do not have dynamic bundles and I have only one hash table per operation instead of per-operation-and-archetype by hashing `(index-of-source-archetype, type-id-of-bundle-or-pair-of-bundles)` with a specialised hasher.

If I did not miss anything which precludes this working at all, then I think these per-operation hash tables would also make sense for hecs. If you are interested in that, I can look into porting that as well. But I also think that should be a separate pull request instead of being folded into this one.

Ralith:
> Do you think the 7.5% improvement is worth the additional complexity?

Hmm, that's not nothing, but unless it's solving a real problem for someone I'm amenable to holding off for now. I also suspect that it'll be easier to factor out shared logic between insert/remove and exchange without a new edge type.

> If I did not miss anything which precludes this working at all, then I think these per-operation hash tables would also make sense for hecs. If you are interested in that, I can look into porting that as well.

Thanks for the interest in all this backporting work! I'm open to this, but it's not immediately obvious to me what the advantage is over per-archetype tables. Maybe open an issue to discuss?

adamreichold:
> I also suspect that it'll be easier to factor out shared logic between insert/remove and exchange without a new edge type.

Yes, that would be my main motivator for this PR as well.

> Thanks for the interest in all this backporting work!

Well, we started from hecs' design and it has helped our simulation work significantly to have a fast yet simple ECS that can be tailored to our needs. So it seems only fair to invest some effort back into hecs.

> I'm open to this, but it's not immediately obvious to me what the advantage is over per-archetype tables. Maybe open an issue to discuss?

I think for now, it is best to get this PR shipshape without an additional edge type.

(The idea behind the per-operation tables is to have fewer separate dynamic allocations and thereby better spatial locality and less memory usage overhead at the cost of slightly more difficult hash table look-ups (which are fast thanks to Hashbrown/SIMD). Producing a serviceable hash out of the type ID which already is a good hash and the small integer archetype indexes seems reasonably simple and efficient by just XOR'ing them.)

Ralith:
> The idea behind the per-operation tables is [...]

Ah, that makes sense. I'd be happy to have that and I imagine it'd be fairly straightforward.

adamreichold:
> Given that it's somewhat fraught unsafe stuff, I'd definitely like to minimize duplication here as much as is feasible. It hasn't needed to be touched in a good while, but when it was fresh it was a source of several serious bugs.

I pushed an updated version that drops the dedicated edges and factors out most of the shared code between `World::{insert,remove,exchange}`. The cuts are a bit weird, but I was unable to factor out something more natural like `World::insert_target` due to how the `InsertTarget` is sometimes owned and borrowed from `World::archetypes`.

(I tried returning something like `Cow<'_, InsertTarget>` or taking `target_storage: &mut Option<InsertTarget>` as an argument, but this both failed due to returning essentially `&'a InsertTarget` which means any borrow from which is derived must last for the whole body of the function which conflicts with borrowing `World::archetypes` several times if the first look-up fails.)

src/world.rs
```diff
@@ -544,23 +544,37 @@ impl World {
...
+        self.insert_bottom_half(entity, components, old_archetype, old_archetype, old_index)
+    }
+
+    fn insert_bottom_half(
+        &mut self,
+        entity: Entity,
+        components: impl DynamicBundle,
+        source_archetype: u32,
+        old_archetype: u32,
+        old_index: u32,
```
Ralith, line 561: Would this be clearer/more concise as `old_loc: Location`?

src/world.rs
```diff
@@ -569,33 +583,34 @@ impl World {
 
         unsafe {
             // Drop the components we're overwriting
-            let source_arch = &mut self.archetypes.archetypes[loc.archetype as usize];
+            let source_arch = &mut self.archetypes.archetypes[old_archetype as usize];
```
Ralith, line 586: It's confusing that `source_arch` is not derived from `source_archetype`. Maybe rename `source_archetype`, perhaps to `archetype_graph_origin` or something similarly explicit?

src/world.rs
```diff
@@ -544,23 +544,37 @@ impl World {
...
+        let old_index = loc.index;
 
+        self.insert_bottom_half(entity, components, old_archetype, old_archetype, old_index)
+    }
+
+    fn insert_bottom_half(
+        &mut self,
+        entity: Entity,
+        components: impl DynamicBundle,
+        source_archetype: u32,
```
Ralith, line 559: We should carefully document the difference between `source_archetype` and `old_archetype`. Would it be correct to say that the former is the associated archetype graph vertex, whereas the latter is the archetype that the components actually reside in? If so, let's also note that these differ only in the case of fused operations, i.e. `exchange`.

src/world.rs
```diff
@@ -544,23 +544,37 @@ impl World {
...
         self.flush();
+
         let loc = self.entities.get_mut(entity)?;
+        let old_archetype = loc.archetype;
+        let old_index = loc.index;
 
+        self.insert_bottom_half(entity, components, old_archetype, old_archetype, old_index)
+    }
+
+    fn insert_bottom_half(
```
Ralith, line 555: nit: `insert_inner`? This is rather more than half of `insert`'s logic.

Ralith (changes requested):
This factoring looks good to me. The important thing is that we're not duplicating fiddly code, and it looks like that's been accomplished in full.

src/world.rs
adamreichold: Push a proposal for an explanatory doc comment.

Ralith (approved):
Thanks! Just needs a rebase.

adamreichold:
> Thanks! Just needs a rebase.

Pushed a rebase which including updates so that the intermediate commits still work.

PR #225, opened 2021-12-26
Use single hash tables for compoment insertion/removal edges to improve spatial locality.

adamreichold:


src/world.rs
```diff
@@ -1245,6 +1246,29 @@ struct InsertTarget {
...
 }
 
+type IndexTypeIdMap<V> = HashMap<(u32, TypeId), V, BuildHasherDefault<IndexTypeIdHasher>>;
+
+#[derive(Default)]
+struct IndexTypeIdHasher(u64);
+
+impl Hasher for IndexTypeIdHasher {
+    fn write_u32(&mut self, index: u32) {
+        self.0 = u64::from(index);
```
Ralith, line 1256: Does std guarantee the order in which tuple fields are hashed? If not, this should probably be xor-assign for robustness; LLVM will probably optimize out the extra work in any case.
```suggestion
        self.0 ^= u64::from(index);
```

Ralith (changes requested):
Very tidy, thanks! Out of curiosity, any changes to the benchmarks? We're probably not exercising cases where this would matter (very high cache pressure, perhaps), but maybe it registers anyway.

src/world.rs
adamreichold: > Does std guarantee the order in which tuple fields are hashed?

For derived `Hash` impls (like those for tuples), fields are always hashed in source order. See https://github.com/rust-lang/rust/pull/91660 for a recent example where this was utilised.

adamreichold: (Actually, the example is not helpful as it relies on the memory layout instead of the hash order.)

Ralith: I don't follow. The linked PR is a custom `Hash` implementation that makes a single `hash` call, rendering ordering questions moot.

Ralith: My goal here is just to not rely on behavior that isn't clearly defined as part of std's public interface, at least if there's any reasonable alternative.

adamreichold:
> Out of curiosity, any changes to the benchmarks? We're probably not exercising cases where this would matter (very high cache pressure, perhaps), but maybe it registers anyway.

I added a benchmark stressing insertion and removal but sticking around a steady state of attached components. This yields
```
test insert_remove                  ... bench:         356 ns/iter (+/- 225)
```
before and
```
test insert_remove                  ... bench:         330 ns/iter (+/- 123)
```
after these changes. I'd say close to noise with a positive tendency?

src/world.rs
adamreichold: I was unable to find a definite statement on the order besides the [source code](https://github.com/rust-lang/rust/blob/673d0db5e393e9c64897005b470bfeb6d5aec61b/compiler/rustc_builtin_macros/src/deriving/hash.rs#L85) itself. Changed as suggested.

Ralith:
Thanks for investigating!

Issue #217, opened 2021-12-02
get_mut2/query_one2 (or similar functionality?)


sdleffler:
There are a few cases where it would be very useful to be able to reference multiple distinct entities w/ potentially the same components, at the same time. The only way this is currently possible is through the `get_unchecked` methods, or with a query which is later filtered out or collected into an auxiliary structure. Unless I've missed something?

Ralith:
That's the intended usecase of `get_unchecked`. A `query_one_unchecked` would be a natural extension. I haven't felt motivated to add safe `...2` versions since in the general case you might need more than 2 and the little bit of unsafe there isn't terrible, but it's not a super strong opinion.

sdleffler:
What about `query_one_n<..., const N: usize>(...) -> [...; N]`?

Ralith:
Ooh, cheeky. I'm not sure how I feel about the implied quadratic assert.

sdleffler:
I mean, if you're looking at the general case over the 2-case, it's gonna be quadratic, and it's no worse w/ just 2 used as the parameter in the same function.

sdleffler:
It would also be really useful to have `get_unchecked` for `ColumnMut`.

Ralith:
> if you're looking at the general case over the 2-case, it's gonna be quadratic

Unless the caller can otherwise prove uniqueness, in which case no check would be necessary. My concern is that an inefficient but flexible and safe method will tend to draw people away from the optimal but unsafe approach. Though now that I type that out, that's probably not such a bad thing in most cases...

I suppose I'm game.

> It would also be really useful to have get_unchecked for ColumnMut.

https://github.com/Ralith/hecs/pull/219

sdleffler:
Working at light speed as always :D

PR #228, opened 2021-12-26
View::get_unchecked

Ralith:


adamreichold:
Hhhmmm, is this really required with `(Prepared)View::get` existing? Or put differently, is it possible to use this soundly where neither `View::get` nor `View::get_mut` would work?

Ralith:
`View::get` cannot be used with queries that yield mutable references, so this is necessary to support cases where you want concurrent multiple access to two different entities from the same query.

adamreichold:
Ah, indeed the references for distinct entities will not overlap. Thanks for explaining! I wonder whether this could be given a safe interface by e.g. passing in `[Entity; N]` and yielding `[Option<QueryItem<'q, Q>>; N]` now that const generics are stable? I guess checking `[Entity; N]` for duplicates would probably end-up `O(N^2)` without additional storage though?

Ralith:
This came up in https://github.com/Ralith/hecs/issues/217. I'm open to the idea; this PR was just the minimum to avoid a regression from the deprecated column API.

PR #231, opened 2021-12-27
RFC: Streamline computation of component removal target archetype

adamreichold:
I think the first commit is almost certainly an improvement as the bundles will usually contain very few types so the binary search almost certainly preferable to the ahash-based `HashSet` in terms of code size and thereby performance. The second one might be pushing it though as the dynamic allocation is done only in the slow path in the first place. The modified method `Bundle::with_static_type_info` might become generally useful in the future though, at the cost of changing the theoretically user-facing API now.

### DIFF
--- a/macros/src/bundle.rs
+++ b/macros/src/bundle.rs
@@ -50,7 +50,7 @@ fn gen_dynamic_bundle_impl(
             }
 
             fn type_info(&self) -> ::std::vec::Vec<::hecs::TypeInfo> {
-                <Self as ::hecs::Bundle>::static_type_info()
+                <Self as ::hecs::Bundle>::with_static_type_info(|info| info.to_vec())
             }
 
             #[allow(clippy::forget_copy)]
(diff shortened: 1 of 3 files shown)


macros/src/bundle.rs
```diff
@@ -50,7 +50,7 @@ fn gen_dynamic_bundle_impl(
             }
 
             fn type_info(&self) -> ::std::vec::Vec<::hecs::TypeInfo> {
```
Ralith, line 52: Should this also become `with_type_info` for symmetry and to permit memory reuse in e.g. `EntityBuilder`?

Ralith (commented):
I think both changes are reasonable. The benefit from the second is probably small, but `Bundle` methods are `#[doc(hidden)]` and hence not considered a stable part of the API, and the tweaked method is consistent with existing patterns and doesn't represent a significant increase in code size or complexity.

macros/src/bundle.rs
adamreichold: I have added another commit doing that change, but I do not immediately see how the memory reuse in `EntityBuilder` would look like?

adamreichold: Do you just want to memoize the collected vectors? Otherwise, I think this is a net loss w.r.t. allocation as it is.

Ralith: My thinking was that `entity_builder::Common` and `CommandBuffer` would own a `Vec<TypeInfo>` that they'd clear and reuse, avoiding reallocs when those structures are reused as recommended. The only time we call this and don't expect a `Vec` out is in `get_insert_target`, so this probably doesn't matter much, but if we can avoid allocating a fresh `Vec` in the implementations of `type_info` then that should save an alloc per insert on average.

adamreichold: > but if we can avoid allocating a fresh Vec in the implementations of type_info then that should save an alloc per insert on average.

But only for the slow path for insertion as otherwise the `InsertTarget` is already cached, right? Whereas in `spawn_inner` we would still need to clone even the re-used `Vec`.

After seeing it implemented, I think I would personally prefer to drop that last commit.

Ralith: Right, that makes sense; let's drop it.

adamreichold: Done.

PR #232, opened 2021-12-27
RFC: Add View::get_mut_n as a safe alterantive to View::get_mut_unchecked.

### LINKED ISSUE #217: get_mut2/query_one2 (or similar functionality?)

adamreichold:
Fixes #217

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -1324,6 +1324,45 @@ impl<'q, Q: Query> View<'q, Q> {
             .as_ref()
             .map(|fetch| fetch.get(meta.location.index as usize))
     }
+
+    /// Like `get_mut`, but allows checked simultaneous access to multiple entities
+    ///
+    /// For N > 3, the check for distinct entities will clone the array and take O(N log N) time.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use hecs::World;
+    /// let mut world = World::new();
+    ///
+    /// let a = world.spawn((1, 1.0));
+    /// let b = world.spawn((2, 4.0));
+    /// let c = world.spawn((3, 9.0));
+    ///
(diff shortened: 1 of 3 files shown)


adamreichold:
`PreparedView::get_mut_n` is missing as I wanted to fix the design first.

Ralith:
Design LGTM.

adamreichold:
Pushed the version including `PreparedView::get_mut_n` and a test case trying to get aliasing mutable references. I also changed the self type to `&mut self` as we still need exclusive access to the view even if the entities are distinct (as we could just call the method twice to produce overlapping exclusive references otherwise).

src/query.rs
```diff
@@ -1394,8 +1433,46 @@ impl<'q, Q: Query> PreparedView<'q, Q> {
             .as_ref()
             .map(|fetch| fetch.get(meta.location.index as usize))
     }
+
+    /// Like `get_mut`, but allows checked simultaneous access to multiple entities
+    ///
+    /// For N > 3, the checking will clone the array and take O(N log N) time.
```
Ralith, line 1439: If we're going to make one of multiple similar methods the canonical one and give it more detailed docs, we should link to it from the non-canonical methods (and probably omit the duplicated details entirely).

Ralith (changes requested):
> I also changed the self type to &mut self

Whoops, good catch!

src/query.rs
adamreichold: Fixed as suggested.

Ralith:
Thanks!

PR #235, opened 2022-01-02
Archetypes are always added, never removed

adamreichold:
So the generation is essentially the number of archetypes.


adamreichold:
I think this also means that the generation could be limited to `u32` as indices of archetypes are stored as `u32`?

Ralith:
Good catch! I think you're correct that we could shrink it, too.

Ralith (approved):

adamreichold:
> I think you're correct that we could shrink it, too.

Added.

PR #247, opened 2022-01-22
WIP: Scoped alternative to World API with granular dynamic borrow checking

kabergstrom:
This PR implements a scoped API that mirrors much of the World API, but does component-level borrow checking at runtime instead of relying on static borrow checking. See ErgoScope. 

Still have a few improvements to do, and I hope to expand the API to include queries, but hopefully I can get some feedback on the design and implementation.


kabergstrom:
I recently realized that with the indirection for data access that exists, it'd be possible to keep component references around between scopes, and just patch up the data pointer on scope exit/enter. Of course, actual access when a scope isn't active would panic. I imagine this could be nice as a way of storing persistent component references in scripts for example. What do you think @Ralith ?

Ralith:
It's an intersecting prospect. Off the cuff, I'm concerned that it would be *very* easy to capture pseudo-references that might be unpredictably invalidated while the scope is inactive. For example, if you have a script/task that gets a component, then waits, then manipulates that component, it would be prone to panicing if the associated entity was despawned for any reason. Would this be too large a footgun?

kabergstrom:
I think we can provide try_read/try_write functions for borrowing, so that errors can be propagated within the scripting environment as opposed to panicking. This would be equivalent to a null reference exception in Unity's C# for example, where the behaviour is similar to what you're talking about. Game objects & components can be destroyed between wait points of coroutines, in which case a null reference exception is thrown on access.

kabergstrom:
I wrote a small simulation to test how it is to use the API

https://github.com/kabergstrom/cogito/blob/master/src/main.rs

kabergstrom:
I think the next step for this feature is to support storing references in the World itself. The current implementation assumes that the number of referenced entities is small at any point in time (allocating a component reference is O(n) to the number of active refs). This has nice cache and memory usage for working with World on the call stack, but will fall apart when used for references between components in the World state. The implementation also has a restriction where it panics if a reference escapes the ErgoScope lifetime. This restriction needs to be lifted in order to be able to store references persistently.

My goal is to make persistent references as ergonomic as the transient references. I will need to implement some form of pointer patching to handle moving and removing components. I think the current indirection-block design is strong, but I think we will need memory compaction to make it more scalable and reliable. Compaction means pointer patching, and at that point we may as well do GC. And that's probably what I'll do - reflection-driven GC.

PR #248, opened 2022-01-29
Introduce method for moving entities between worlds

Ralith:


src/take.rs
```diff
@@ -0,0 +1,52 @@
+use alloc::vec::Vec;
+use core::mem;
+
+use crate::{Archetype, DynamicBundle, TypeInfo};
+
+/// An entity removed from a `World`
+pub struct TakenEntity<'a> {
+    archetype: &'a Archetype,
```
adamreichold, line 8: I think a `&'a mut Archetype` would be better documenting the underlying assumptions here. It already is essential that as `TakenEntity<'a>` is produced by a `&'a mut World` and I don't think this is straight foward to generalize to taken multiple entities as there is only one "one past the end" slot per archetype.

src/world.rs
```diff
@@ -878,6 +881,18 @@ impl World {
         ColumnMut::new(entities, archetypes, PhantomData)
     }
 
+    /// Despawn `entity`, yielding a [`DynamicBundle`] of its components
+    ///
+    /// Useful for moving entities between worlds.
+    pub fn take(&mut self, entity: Entity) -> Result<TakenEntity<'_>, NoSuchEntity> {
+        let loc = self.entities.free(entity)?;
```
adamreichold, line 888: `World::despawn` calls `World::flush` before doing anything. Is that unnecessary here?

src/archetype.rs
```diff
@@ -26,6 +26,7 @@ use crate::{Access, Component, Query};
 /// [`World`](crate::World).
 pub struct Archetype {
     types: Vec<TypeInfo>,
+    type_ids: Box<[TypeId]>,
```
adamreichold, line 29: This seems a bit of a price to pay to produce a matching slice of `TypeId`. Maybe the bundle interface can actually make due with `(ExactSize)Iterator` so that `Archetype::types` could also fulfil the requirement?

adamreichold (commented):
Having this infrastructure available, I wonder whether replacing some of the manual memory management of the `CommandBuffer` by just using a `World` in its implementation would be useful? Mostly to make it easier to implement more features on `CommandBuffer`. (I am thinking of things like encoding the operation using private component in that stash world and then querying for them when the buffer is applied.)

src/take.rs
adamreichold: Since this basically has exclusive access to `World` and thereby `Archetype`, I think it could also be more efficient by avoiding the preparatory swap with the last entity stored in the source archetype? (Thereby going from 3+1 copies to 1+1, i.e. entity to temp, last entity to entity, temp to last entity and last entity to destination versus entity to destination and last entity to entity.)

Ralith: > I think a &'a mut Archetype would be better documenting the underlying assumptions here

I agree. Unfortunately, a mutable reference can't be moved out of `Self`, which means we can't `mem::forget(self)` before moving components out of `archetype` in `put`, which means a panic inside `put`'s callback might cause `Drop` to be invoked on values which have already been moved away, which is UB. We could work around this by storing the reference in `Option`, but that consumes the niche, which seems unfortunate.

> it could also be more efficient by avoiding the preparatory swap with the last entity stored in the source archetype?

The current design is motivated by maintaining good behavior even if `TakenEntity` is leaked. However, on review I think we can still preserve that by deferring world changes to `put`. I'm kind of curious exactly how much worse `swap_nonoverlapping` is than `copy_nonoverlapping`, though, since avoiding the former is what this proposal boils down to.

src/world.rs
Ralith: Good catch; absolutely necessary.

src/archetype.rs
Ralith: My feeling is that incremental fixed costs on archetype construction are not a significant overhead, since archetype construction is already relatively expensive.

`with_ids` is only called in connection with `ArchetypeSet::get`, which passes it to `HashMap::<Box<[TypeId]>, _>::get`. Substituting an iterator there is awkward, likely requiring replacing the `HashMap` with a [`RawTable`](https://docs.rs/hashbrown/latest/hashbrown/raw/struct.RawTable.html).

Given the uncertain cost and complexity of mitigating it, I am inclined to deem this a premature opt.

Ralith:
> I wonder whether replacing some of the manual memory management of the CommandBuffer by just using a World in its implementation would be useful?

This is an interesting idea. Merging archetypes is also liable to be significantly more efficient than inserting components one at a type from row-major storage, in effect moving cost to `CommandBuffer` population time, which is comparatively easy to offload--not that this is a cost anyone has complained about yet. On the other hand, the only commands I would think to add at present are despawn and remove operations, which are straightforward to express with simpler means.

src/take.rs
Ralith: Revised. The result is a slightly larger `TakenEntity`, but simpler code overall, which I'm happy with.

src/archetype.rs
adamreichold: > Given the uncertain cost and complexity of mitigating it, I am inclined to deem this a premature opt.

Agreed. I was not thinking far enough and using `Iterator` would just mess things up elsewhere.

One idea related to bundles even if unrelated to this PR would be to modify the `DynamicBundle` trait to be more operation oriented, e.g. instead of providing slices of `TypeId` or `TypeInfo`, it could get methods for performing more specific operations like "check whether this bundle contains `T`?" or "here is a set of type ID, insert all type ID which are part of you but not yet part of the set". This would probably make implementing the trait more involved, but also more flexible as to the shape of the data backing those operations.

src/take.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+impl Drop for TakenEntity<'_> {
+    fn drop(&mut self) {
+        unsafe {
+            if let Some(moved) = self.archetype.remove(self.index, self.drop) {
+                self.entities.meta[moved as usize].location.index = self.index;
+            }
+            self.entities.free(self.entity).unwrap();
```
adamreichold, line 61: Does freeing the entity need to be delayed until here or is this so that a leaked `TakenEntity` is a no-op? (Freeing the entity up front would also reduce the size of `TakenEntity`.)

Ralith: If we free the entity ID early, then if the `TakenEntity` is leaked we are left in an inconsistent state wherein an `Entity` appears not to exist when accessed directly but still appears in queries. This would be *very* confusing so I think we should take care to prevent it.

src/archetype.rs
Ralith: That's an intriguing prospect, though indeed best pursued in future work. It might help rein in the complexity/ugliness of the traits, perhaps even to the point that we could make them public. Some additional `World` and/or `Archetype` APIs would likely need to be defined for implementations to call into.

Ralith: Opened https://github.com/Ralith/hecs/issues/249 to track that.

src/take.rs
adamreichold: Right. I was thinking only of `World::get`, but queries would still visit the entity.

Ralith:
Thanks for the great review and discussion, @adamreichold!

PR #259, opened 2022-03-22
Clarify that using World::clear will lead to entity ID repeating.

adamreichold:
We ran into this during modelling and were confused, so thought it would be nice to at least clarify this in the documentation.

(Fixing it would most likely make `World::clear` significantly more costly to implement as all generations would need to be bumped, ideally except for those already on the free list.)

cc @mlange-42

### DIFF
--- a/src/world.rs
+++ b/src/world.rs
@@ -341,7 +341,7 @@ impl World {
 
     /// Despawn all entities
     ///
-    /// Preserves allocated storage for reuse.
+    /// Preserves allocated storage for reuse but clears metadata so that [`Entity`] values will repeat (in contrast to [`despawn`][Self::despawn]).
     pub fn clear(&mut self) {
         for x in &mut self.archetypes.archetypes {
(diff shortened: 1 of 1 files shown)


src/world.rs
```diff
@@ -341,7 +341,7 @@ impl World {
 
     /// Despawn all entities
     ///
-    /// Preserves allocated storage for reuse.
+    /// Preserves allocated storage for reuse but clears metadata so that [entity ID][Entity] will repeat (in contrast to [`despawn`][Self::despawn]).
```
Ralith, line 344:
```suggestion
    /// Preserves allocated storage for reuse but clears metadata so that [`Entity`][Entity] values will repeat (in contrast to [`despawn`][Self::despawn]).
```

Ralith (changes requested):
Thanks! This does seem like a possible footgun.

Ralith (approved):

PR #261, opened 2022-05-05
Pass serializers by value to serializer context traits

Ralith:
Gives more flexibility to implementers, allowing e.g. cheeky type erasure tricks.


LPGhatguy:
This PR looks great, it enables gluing `SerializeContext` into erased-serde very cleanly!

PR #263, opened 2022-05-16
impl Send and Sync for View

Uriopass:
Borrow checking is done at .query() time. Safe to send/share between threads as there is no interior mutability here.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -684,8 +684,8 @@ impl<'w, Q: Query> QueryBorrow<'w, Q> {
     }
 }
 
-unsafe impl<'w, Q: Query> Send for QueryBorrow<'w, Q> {}
-unsafe impl<'w, Q: Query> Sync for QueryBorrow<'w, Q> {}
+unsafe impl<'w, Q: Query> Send for QueryBorrow<'w, Q> where <Q::Fetch as Fetch<'w>>::Item: Send {}
+unsafe impl<'w, Q: Query> Sync for QueryBorrow<'w, Q> where <Q::Fetch as Fetch<'w>>::Item: Send {}
 
 impl<'w, Q: Query> Drop for QueryBorrow<'w, Q> {
     fn drop(&mut self) {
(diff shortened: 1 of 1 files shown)


src/query.rs
```diff
@@ -1269,6 +1269,9 @@ pub struct View<'q, Q: Query> {
     fetch: Vec<Option<Q::Fetch>>,
 }
 
+unsafe impl<Q: Query> Send for View<'_, Q> {}
```
adamreichold, line 1272: I think both impls need to be conditional on
```rust
<Q::Fetch as Fetch<'q>>::Item: Send
```
as otherwise values `T: !Sync` could be accessed from multiple threads if e.g. `Item == &'q T`.

Uriopass: I think it's ok because components need to be Send + Sync

https://github.com/Ralith/hecs/blob/master/src/world.rs#L1025

A lot of code would be broken otherwise :D

Uriopass: Updated the other bounds too

adamreichold: > I think it's ok because components need to be Send + Sync

Right, I forgot about that. I do prefer the explicit bounds here though to enable "local reasoning."

> Updated the other bounds too

I think `Item: Send` should suffice for `View: Sync` etc. as access to `&View` will always yield `Item`, never `&Item`.

Ralith: Would it make more sense to constrain `Item` (and perhaps `State`) in the definition of `Fetch`?

adamreichold: > Would it make more sense to constrain Item (and perhaps State) in the definition of Fetch?

I don't there is anything in `Fetch` that requires `Send` or `Sync`? Everything in `mod query` can work without it, except when queries etc. need to parallelized or sent to/accessed from other threads.

IMHO, I think either the bounds proposed here are preferable due to explicitness or relying on the bounds on `Component` is preferable due to convenience.

PR #266, opened 2022-05-24
Fix spawn_at creating phantom entities

### LINKED ISSUE #264: `spawn_at` is panicking
When I try running the following code:
```
pub fn rehydrate(&mut self, entity: Entity) {
    if !self.world.contains(entity) {
      self.world.spawn_at(entity, ());
    }
}
```

Rust panics with this message:
`panicked at 'index out of bounds: the len is 64 but the index is 4294967294'`

It still manages to create a few entities before doing so, though:
```
INFO engine/src/application/scene.rs:38 Spawning 0v1
INFO engine/src/application/scene.rs:38 Spawning 1v1
INFO engine/src/application/scene.rs:38 Spawning 2v1
INFO engine/src/application/scene.rs:38 Spawning 3v1
INFO engine/src/application/scene.rs:38 Spawning 5v1
INFO engine/src/application/scene.rs:38 Spawning 7v1
INFO engine/src/application/scene.rs:38 Spawning 9v1
```

Ralith:
Fixes #264.

### DIFF
--- a/src/entities.rs
+++ b/src/entities.rs
@@ -370,7 +370,7 @@ impl Entities {
         self.verify_flushed();
 
         let meta = self.meta.get_mut(entity.id as usize).ok_or(NoSuchEntity)?;
-        if meta.generation != entity.generation {
+        if meta.generation != entity.generation || meta.location.index == u32::MAX {
             return Err(NoSuchEntity);
         }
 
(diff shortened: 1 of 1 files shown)


src/entities.rs
```diff
@@ -400,10 +400,13 @@ impl Entities {
...
 
     pub fn contains(&self, entity: Entity) -> bool {
-        // Note that out-of-range IDs are considered to be "contained" because
-        // they must be reserved IDs that we haven't flushed yet.
         match self.meta.get(entity.id as usize) {
-            Some(meta) => meta.generation == entity.generation,
+            Some(meta) => {
+                meta.generation == entity.generation
+                    && (meta.location.index != u32::MAX
+                        || self.pending[self.free_cursor.load(Ordering::Relaxed).max(0) as usize..]
```
adamreichold, line 407: This looks heavyweight from a performance perspective. Is this unlikely to be hit?

Ralith: Short-circuiting ensures it only arises on reserved or "phantom" entities. Further, I don't think `contains` is typically hot-loop for people, although that's speculative. We could also drop it and just redefine the semantics of `contains` to exclude reserved but unflushed entities; that seems like it would generally be fine, but on the other hand it'd be difficult to emulate externally if someone does care about that case.

adamreichold: > We could also drop it and just redefine the semantics of contains to exclude reserved but unflushed entities;

I think its better to wait if people complain about the performance than the weaken the semantics.

PR #270, opened 2022-06-21
fix typo in ffa_simulation.rs enity -> entity

jrmiller82:
Fix a minor typo in an example file: `enity` -> `entity`


Ralith (approved):
Thanks!

PR #274, opened 2022-07-11
Refactor World::get to take reference types

Ralith:
Many users have reported confusion arising from the use of component types in `World::get` in contrast to the use of reference sin `World::query` and friends. This replaces `get` and `get_mut` on `World` and `EntityRef` with a single `get` that is generic over *references to components*, so that the intuitive thing Just Works, and passing an explicit component type generates a trait error.

For consistency, `EntityBuilder:;get` and `get_mut` are both refactored similarly, although they cannot be combined into a single method as they do not perform dynamic borrow checking.

This also resolves the inconsistency between `get_mut` referring to the mutability of the component borrow, while other `_mut` methods on `World` instead refer to the mutability of the world borrow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+### Added
+- `World::satisfies` and `EntityRef::satisfies` to check if an entity would match a query
+
 ### Changed
 - Many generic methods that previously took a `Component` now instead take either a
   `ComponentRef<'a>` or a `Query` to improve consistency with queries and address a common footgun:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
     component types
   - `EntityBuilder` and `EntityBuilderClone`'s `get` and `get_mut` refactored along the same lines
     for consistency
+  - The `With`/`Without` query transformers now take a query that entities must/mustn't match rather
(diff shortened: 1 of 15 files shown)


adamreichold:
I have not formed an opinion of whether this simplifies explaining things (In my mental model, `World::get(_mut)` were the primitives out of which the queries are built via their DSL.), I wanted to add that the query combinators potentially have a similar foot gun, e.g. `Query<With<(&R, &S), &T>>` does not work as one might expect and the correct `Query<With<(&R, &S), T>>` might be surprising. (I guess it isn't completely the same as `&T` and `&mut T` imply no difference for the query combinator.)

Ralith:
I've received many independent user reports of footgun-related injuries here, so I'm convinced of the benefit.

`With`/`Without` are an interesting point. I've also been a bit unhappy with how they have to be nested to consider multiple components. Maybe we should refactor those to operate on two queries, rather than a query and a component type? Logically equivalent to the 2-tuple combinator, except that the results for the second item are not yielded (or borrowed).

adamreichold:
> Maybe we should refactor those to operate on two queries, rather than a query and a component type?

This sounds sensible and would also be nicely consistent with `Satisfies`.

Ralith:
Comically, I have just now discovered a bug in my own project that this would have prevented. Will probably publish a release soon.

PR #281, opened 2022-08-14
Implement into_iter_batched for QueryMut

Uriopass:
Skips the borrow checking

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -824,6 +824,13 @@ impl<'q, Q: Query> QueryMut<'q, Q> {
             iter: unsafe { QueryIter::new(self.iter.meta, self.iter.archetypes) },
         }
     }
+
+    /// Like `into_iter`, but returns child iterators of at most `batch_size` elements
+    ///
+    /// Useful for distributing work over a threadpool.
+    pub fn into_iter_batched(self, batch_size: u32) -> BatchedIter<'q, Q> {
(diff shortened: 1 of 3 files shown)


Uriopass:
The CI test failing talked about an error message unrelated to the PR that looked like it changed in the latest Rust, fixed it in [another commit](https://github.com/Ralith/hecs/pull/281/commits/11dbef93655c345cf8759d528372dc195b5cecc9).

src/query.rs
```diff
@@ -824,6 +824,13 @@ impl<'q, Q: Query> QueryMut<'q, Q> {
             iter: unsafe { QueryIter::new(self.iter.meta, self.iter.archetypes) },
         }
     }
+
+    /// Like `into_iter`, but returns child iterators of at most `batch_size` elements
+    ///
+    /// Useful for distributing work over a threadpool.
+    pub fn iter_batched(self, batch_size: u32) -> BatchedIter<'q, Q> {
```
Ralith, line 831: Since this takes ownership of `self`:
```suggestion
    pub fn into_iter_batched(self, batch_size: u32) -> BatchedIter<'q, Q> {
```

Ralith (changes requested):
Thanks!

src/query.rs
Uriopass: Done!

Ralith (approved):

PR #282, opened 2022-09-12
Add `QueryIter::peek`

Ralith:


src/query.rs
```diff
@@ -731,6 +731,13 @@ impl<'q, Q: Query> QueryIter<'q, Q> {
             iter: ChunkIter::empty(),
         }
     }
+
+    /// Access the next result without advancing the iterator
+    pub fn peek(&mut self) -> Option<<Q::Fetch as Fetch<'_>>::Item> {
+        // Safety: returned lifetime borrows `self`, ensuring it can't overlap with a future call to
+        // `peek` or `next`
+        unsafe { self.iter.peek().map(|x| x.1) }
```
adamreichold, line 739: Why strip the entity ID here if it is returned by the `Iterator` impl? Isn't that something the caller could need and if not ignore?

Ralith: Yeah, that should definitely be included.

PR #286, opened 2022-10-15
Add removal of `World::get_mut` and `EntityRef::get_mut` to the changelog

### LINKED ISSUE #284: Removal of `World::get_mut` is not mentioned in the changelog
[`World::get_mut`](https://docs.rs/hecs/0.7.7/hecs/struct.World.html#method.get_mut) was removed in hecs 0.8.0, presumably because [`World::get` can now replace it](https://github.com/Ralith/hecs/blob/master/CHANGELOG.md#080). However, this is not explicitly mentioned in the changelog, which can cause confusion during upgrades.

PatchMixolydic:
These methods were removed in hecs 0.8.0 (in #274), but previously, their removal wasn't explicitly mentioned in the changelog. This commit mentions them by name, which should make the upgrade path to `World::get`/`EntityRef::get` a bit clearer.

Fixes #284.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,9 +27,12 @@
 - Many generic methods that previously took a `Component` now instead take either a
   `ComponentRef<'a>` or a `Query` to improve consistency with queries and address a common footgun:
   - `World::get`, `EntityRef::get`, and `Archetype::get` now take shared or unique references to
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Thanks!

PR #293, opened 2023-01-11
implemented `has`

sanbox-irl:
Added `has` and `has_static`. I went with `has_static` instead of `static_has` for the sake of autocompletion, but not sure if you have a preference.

Moreover, adjusted documentation to make these two functions a public member of the interface.

### DIFF
--- a/src/bundle.rs
+++ b/src/bundle.rs
@@ -59,19 +59,6 @@ pub unsafe trait DynamicBundle {
 /// interface of this trait, except `has_static`, is a private implementation detail.
 #[allow(clippy::missing_safety_doc)]
 pub unsafe trait Bundle: DynamicBundle {
-    /// Checks if the Bundle contains the given `T`:
-    ///
-    /// ```
-    /// fn test_bundle<T: hecs::Bundle>(input: T) {
-    ///     assert!(T::has_static::<i32>());
-    ///     assert!(T::has_static::<f32>());
-    ///     assert!(!T::has_static::<usize>());
-    /// }
-    /// ```
-    fn has_static<T: Component>() -> bool {
-        Self::with_static_ids(|types| types.contains(&TypeId::of::<T>()))
-    }
-
     #[doc(hidden)]
     fn with_static_ids<T>(f: impl FnOnce(&[TypeId]) -> T) -> T;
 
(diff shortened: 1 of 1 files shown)


src/bundle.rs
```diff
@@ -42,9 +56,22 @@ pub unsafe trait DynamicBundle {
...
+    /// Checks if the Bundle contains the given `T`:
+    ///
+    /// ```
+    /// fn test_bundle<T: hecs::Bundle>(input: T) {
+    ///     assert!(T::has_static::<i32>());
+    ///     assert!(T::has_static::<f32>());
+    ///     assert!(!T::has_static::<usize>());
+    /// }
+    /// ```
+    fn has_static<T: Component>() -> bool {
```
Ralith, line 71: What's the use case for this?

src/bundle.rs
```diff
@@ -25,6 +25,20 @@ pub unsafe trait DynamicBundle {
...
+    ///
+    /// ```
+    /// # use hecs::DynamicBundle;
+    ///
+    /// let my_bundle = (0i32, 10.0f32);
+    /// assert!(my_bundle.has::<i32>());
+    /// assert!(my_bundle.has::<f32>());
+    /// assert!(!my_bundle.has::<usize>());
+    /// ```
+    fn has<T: Component>(&self) -> bool {
```
Ralith, line 38: `EntityBuilder`-related bundles could probably have a more efficient override for this.

src/bundle.rs
sanbox-irl: in my own case, I can use this directly rather than `has`, but probably it ends up all the same after the optimizer. I haven't checked yet

src/bundle.rs
sanbox-irl: Want me to take a crack at implementing that, or are you asking me to not provide a default implementation?

Ralith: Now that you mention it, implementing this directly for each type rather than with this default might not be a bad idea, since it lets us skip the sorting in the type ID accessors.

src/bundle.rs
Ralith: The static variants of the other methods are motivated by cases where there's no `self` to call on (e.g. `remove`). Maybe that makes sense here too, but it bears some consideration if it's not part of your use case.

Could we name this `has` as well, or would that have worse ergonomics in practice?

sanbox-irl: Which doesn't describe any of my use cases. I'd be for removing this method (which incidentally, yeah, has bad ergonomics with the name `has`) until such time as its needed by anyone then

src/bundle.rs
```diff
@@ -42,7 +56,7 @@ pub unsafe trait DynamicBundle {
 /// A statically typed collection of components
 ///
 /// Bundles composed of exactly the same types are semantically equivalent, regardless of order. The
-/// interface of this trait is a private implementation detail.
+/// interface of this trait, except `has_static`, is a private implementation detail.
```
Ralith, line 59: The change to this comment is outdated.

Ralith:
Merged via https://github.com/Ralith/hecs/pull/294.

PR #295, opened 2023-01-16
clippy lints fixed

sanbox-irl:
Fixed the various lints clippy throws right now. Some of these are uglier fixes, but it makes working in this repo nicer by reducing warnings.

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -331,12 +331,13 @@ fn build_builder_clone() {
 }
 
 #[test]
+#[allow(clippy::redundant_clone)]
 fn cloned_builder() {
     let mut builder = EntityBuilderClone::new();
     builder.add(String::from("abc")).add(123);
 
     let mut world = World::new();
-    let e = world.spawn(&builder.build());
+    let e = world.spawn(&builder.build().clone());
     assert_eq!(*world.get::<&String>(e).unwrap(), "abc");
     assert_eq!(*world.get::<&i32>(e).unwrap(), 123);
 }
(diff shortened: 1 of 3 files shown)


sanbox-irl:
fwiw, if the maintainers would like to just reject this, no hard feelings. I don't feel strongly about clippy, but some of these are reasonable to fix. Notably the "push_str" things

tests/tests.rs
```diff
@@ -336,7 +336,7 @@ fn cloned_builder() {
     builder.add(String::from("abc")).add(123);
 
     let mut world = World::new();
-    let e = world.spawn(&builder.build().clone());
+    let e = world.spawn(&builder.build());
```
Ralith, line 339: This defeats the point of the test.

Ralith (changes requested):
These mostly seem reasonable; thanks!

tests/tests.rs
sanbox-irl: fixed, but forced into an explicit clippy allow, which I find ugly (i like repos to magically pass clippy lints but otherwise not have any obvious signs that they're aware of clippy, since otherwise, of course, you're just trading noise of one kind for noise of another).

Alternative that is just weirder:
```
    let e = world.spawn(&Clone::clone(&builder.build()));
```
and that wouldn't require a clippy allow.

That's less a methodology and more clippy just doesn't catch it, so probably not future proof

Ralith (approved):

tests/tests.rs
Ralith: I think this is a fine place to explicitly allow, since it's a pattern that would be dumb anywhere *but* a test of `Clone`'s correctness.

PR #301, opened 2023-03-10
Fix World::{query_one, query_one_mut} violating unique borrow.

### LINKED ISSUE #300: World::query_one(_mut) is unsound
Given the following example, it is possible to be given an `&mut` reference of the same value twice.

An example (this will need to be run a few times to exhibit the race condition):
```rust
use std::sync::{Barrier, Arc};

use hecs::World;

let mut world = World::new();
let entity = world.spawn(255u32);

// query_one_mut here is unsound because it returns multiple exclusive references to the `u32` inserted above.
let (value, again) = world.query_one_mut::<(&mut u32, &mut u32)>(entity).unwrap();

// A barrier is used to ensure that the two threads will race each other.
let barrier = Arc::new(Barrier::new(2));
let b1 = Arc::clone(&barrier);
let b2 = Arc::clone(&barrier);

std::thread::scope(|scope| {
    let handle1 = scope.spawn(|| {
        b1.wait();
        *value = 300;
    });

    let handle2 = scope.spawn(|| {
        b2.wait();
        *again = 400;
    });

    handle1.join().unwrap();
    handle2.join().unwrap();
});

assert_eq!(*value, 300);
```

i509VCB:
Fixes #300

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - `Archetype::satisfies` to check whether an archetype satisfies a query
 - `serialize::{row,column}::serialize_satisfying` to serialize all entities that satisfy a query
 
+### Fixed
+-  `World::{query_one, query_one_mut}` allowed violating a unique borrow.
+
 # 0.9
 
 ### Changed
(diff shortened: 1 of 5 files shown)


Ralith (approved):
Thanks!

i509VCB:
Had to force push to fix lint.

Ralith:
Released in 0.10.

PR #307, opened 2023-03-25
Implement updating components in command buffers

cjhowedev:
Adds the ability to command buffers to update entities' components. This can be used to implement operations on components such as incrementing a counter or appending to an array.

### DIFF
--- a/src/command_buffer.rs
+++ b/src/command_buffer.rs
@@ -13,7 +13,6 @@ use alloc::boxed::Box;
 use crate::alloc::alloc::{alloc, dealloc, Layout};
 use crate::alloc::vec::Vec;
 use crate::archetype::TypeInfo;
-use crate::Query;
 use crate::{align, DynamicBundle};
 use crate::{Bundle, Entity};
 use crate::{Component, World};
@@ -34,7 +33,7 @@ use crate::{Component, World};
 pub struct CommandBuffer {
     entities: Vec<EntityIndex>,
     remove_comps: Vec<RemovedComps>,
-    update_comps: Vec<UpdatedComps>,
+    commands: Vec<Command>,
     despawn_ent: Vec<Entity>,
     storage: NonNull<u8>,
     layout: Layout,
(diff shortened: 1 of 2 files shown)


cjhowedev:
I think that maybe we can do better than this API. It's very simple, and it turns out I need an upsert to implement the kinds of commands I want to support.

cjhowedev:
Okay, I ended up with a greatly simplified API which supports all prior use cases. Let me know what you think!

src/command_buffer.rs
```diff
@@ -169,6 +177,10 @@ impl CommandBuffer {
             let _ = world.despawn(*entity);
         }
 
+        for comp in self.commands.iter() {
```
adamreichold, line 180: If this would use `self.commands.drain(..)`, then `Command` could become an `FnOnce` and consume its captures which would increase the usefulness IMHO.

src/command_buffer.rs
```diff
@@ -120,6 +123,11 @@ impl CommandBuffer {
         self.remove::<(T,)>(ent);
     }
 
+    /// Updates the world with the given command after all other operations have run
+    pub fn command<F: Fn(&mut World) + Send + Sync + 'static>(&mut self, command_fn: F) {
```
adamreichold, line 127: I wonder if this needs to be generic and pay for the associated code size overhead? Maybe this could take `Command` and that could have a

```rust
impl From<F> for Command where F: FnOnce(&mut World) + Send + Sync + 'static
```

to keep the generic as limited as possible? Or does this hurt the ergonomics of using it too much?

Ralith: As a one-liner, this being generic isn't going to impose a significant cost. A generic `From` impl would also have the same cost.

Ralith:
Thanks for the PR!

I'm unsure if this is a direction we should go in. In a game (for example) hecs is not intended to be the entirety of your world representation; in practice applications might be better served by maintaining their own `Vec` of closures which accept a more all-encompassing state struct that includes non-ECS data. By providing a narrower version, we risk promoting a pattern that may prove awkwardly limiting. Meanwhile, implementing this pattern downstream is simple and no less efficient.

adamreichold:
> in practice applications might be better served by maintaining their own Vec of closures

I would also suggest to refrain from this. In my personal experience, one can also be both more readable and more efficient by not storing closures, but rather the data (entity ID, new or updated values, etc.) a uniform set of closures would operate on because the code itself is usually fixed at least on a per-system basis.

PR #314, opened 2023-04-06
Add `Ref::map` and `RefMut::map`.

burtonageo:
This commit adds analogues of [std::cell::Ref::map](https://doc.rust-lang.org/stable/std/cell/struct.Ref.html#method.map) and [std::cell::RefMut::map](https://doc.rust-lang.org/stable/std/cell/struct.RefMut.html#method.map), which allow the contents of a `Ref`/`RefMut` to be transformed without reborrowing. This can make it easier to write certain kinds of functions which return data borrowed from components.

An example of this kind of code is below:

```rust
trait MyComponentTrait {
    fn get_state(&self) -> &dyn MyState;
}

fn entity_get_state<T: MyComponentTrait>(ent: EntityRef<'_>) -> Option<Ref<'_, dyn MyState>> {
    let component = ent.get::<&'_ T>();
    component.map(|comp_ref| Ref::map(comp_ref, |comp| comp.get_state()))
}

struct MyComponentInstance {
    // state unspecified...
}

impl MyComponentTrait for MyComponmentInstance {
    // implementation unspecified...
}

let mut comp_ty_to_state_getter: HashMap<TypeId, fn(EntityRef<'_>) -> Option<Ref<'_, dyn MyState>> = Default::default();
comp_ty_to_state_getter.insert(TypeId::of::<MyComponmentInstance>(), entity_get_state::<MyComponmentInstance>);
```

From there, when you have an entity, you could iterate over the `TypeId`s of the components in that entity, and then
call use the `comp_ty_to_state_getter` to call common functionality on all of them. Examples of where this could
be useful include serialisation, or visiting component fields to render them in a GUI inspector.

As part of this change, the sizes of the `Ref` and `RefMut` types have increased to store the `TypeId` of the original
component, and the `Archetype` can now be released polymorphically using that `TypeId`.

Let me know if there is anything else I should change or if there are any tests I should add.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ### Added
 - `Ref::map` and `RefMut::map` to reborrow a subfield of a component, or cast it to an unsized type
 
+### Fixed
+- `Ref::clone()` is now callable when containing types which do not implement `Clone`
+- `Ref::clone()` could cause a panic as internal state wasn't being updated correctly
+
 # 0.10.2
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+### Added
+- `Ref::map` and `RefMut::map` to reborrow a subfield of a component, or cast it to an unsized type
+
 # 0.10.2
 
 ### Fixed
--- a/src/entity_ref.rs
+++ b/src/entity_ref.rs
@@ -1,6 +1,7 @@
 use core::any::TypeId;
 use core::fmt::{self, Debug, Formatter};
-use core::ops::{Deref, DerefMut};
+use core::marker::PhantomData;
+use core::ops::{Deref, DerefMut, FnOnce};
 use core::ptr::NonNull;
 
 use crate::archetype::Archetype;
@@ -109,12 +110,10 @@ unsafe impl<'a> Send for EntityRef<'a> {}
 unsafe impl<'a> Sync for EntityRef<'a> {}
 
 /// Shared borrow of an entity's component
-#[derive(Clone)]
-pub struct Ref<'a, T: Component> {
-    archetype: &'a Archetype,
-    /// State index for `T` in `archetype`
-    state: usize,
+pub struct Ref<'a, T: ?Sized> {
+    borrow: ComponentBorrow<'a>,
     target: NonNull<T>,
+    _phantom: PhantomData<&'a T>,
 }
 
 impl<'a, T: Component> Ref<'a, T> {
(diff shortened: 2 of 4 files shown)


Ralith:
Thanks for the PR! I agree that this API is well motivated. Because it should be impossible to construct a `Ref` or `RefMut` with an incorrect `state` field wrt. its `archetype`, I think we should skip the extra `TypeId` and instead add an unsafe internal API to `Archetype` to borrow/release by state index.

src/entity_ref.rs
```diff
@@ -135,26 +136,64 @@ impl<'a, T: Component> Ref<'a, T> {
     }
 }
 
-unsafe impl<T: Component> Send for Ref<'_, T> {}
-unsafe impl<T: Component> Sync for Ref<'_, T> {}
+unsafe impl<T: ?Sized> Send for Ref<'_, T> {}
```
Ralith, line 139:
```suggestion
unsafe impl<T: ?Sized + Sync> Send for Ref<'_, T> {}
```

burtonageo:
Done

src/entity_ref.rs
```diff
@@ -135,26 +136,64 @@ impl<'a, T: Component> Ref<'a, T> {
...
+unsafe impl<T: ?Sized> Sync for Ref<'_, T> {}
```
Ralith, line 140:
```suggestion
unsafe impl<T: ?Sized + Sync> Sync for Ref<'_, T> {}
```

src/entity_ref.rs
```diff
@@ -178,23 +217,62 @@ impl<'a, T: Component> RefMut<'a, T> {
     }
 }
 
-unsafe impl<T: Component> Send for RefMut<'_, T> {}
-unsafe impl<T: Component> Sync for RefMut<'_, T> {}
+unsafe impl<T: ?Sized> Send for RefMut<'_, T> {}
+unsafe impl<T: ?Sized> Sync for RefMut<'_, T> {}
```
Ralith, line 221: As above

Ralith (changes requested):
nit: these aren't required for soundness under the current API, but might avoid future bugs/breakage.

src/entity_ref.rs
```diff
@@ -135,26 +136,64 @@ impl<'a, T: Component> Ref<'a, T> {
     }
 }
 
-unsafe impl<T: Component> Send for Ref<'_, T> {}
-unsafe impl<T: Component> Sync for Ref<'_, T> {}
+unsafe impl<T: ?Sized + Sync> Send for Ref<'_, T> {}
```
burtonageo, line 139: I've marked the `Send` impl for `Ref` as requiring its wrapped value to be `Sync`, as it is easy to clone `Ref`s and alias the underlying value, so it is only safe to send a `Ref` across a thread if `T` is `Sync`.

src/entity_ref.rs
```diff
@@ -178,23 +217,62 @@ impl<'a, T: Component> RefMut<'a, T> {
     }
 }
 
-unsafe impl<T: Component> Send for RefMut<'_, T> {}
-unsafe impl<T: Component> Sync for RefMut<'_, T> {}
+unsafe impl<T: ?Sized + Send> Send for RefMut<'_, T> {}
```
burtonageo, line 220: As the `RefMut` provides exclusive access, I've marked the `Send` impl as just requiring the `T` to be `Send`, but I'm not sure about this :-/

burtonageo:
I'd also like to rebase this onto https://github.com/Ralith/hecs/pull/315 when that's approved, as the `Debug` impls could be expanded to require `?Sized + Debug`

src/entity_ref.rs
Ralith: `RefMut` also provides shared access via `Deref`, so it should be possible to usefully share a `&RefMut<T>` between threads iff `T: Sync`.

Ralith: Er, but this is the `Send` impl. I think this is correct.

Ralith (approved):

Ralith:
I like the refactoring, but bundling it together with the API extension will make this take longer to review. You might also want to rebase to fix the conflict.

burtonageo:
Rebased!

Sorry about the extra refactoring. I decided to add a test, which exposed a bug in the impl of `Ref::clone()`, where A. The `#[derive(Clone)]` was putting an extra requirement that `impl<T: Clone> Clone for Ref<T>` which was unnecessary, and also not incrementing the borrow ref counts in the archetype for immutable borrows of the component.

Therefore, I decided to refactor the borrowing permissions into the `ComponentBorrow` and `ComponentBorrowMut` types, so that these could manage the borrowing lifetime from the `Archetype`, and then the `Ref` and `RefMut` types could focus on being useful wrappers over the `NonNull<T>`. This fixed the case in test where the extra clone of the `Ref<T>` would cause a panic due to imbalanced references, and it also simplified the `{Ref, RefMut}::map` implementations, as the borrow could be passed to the new `Ref`, rather than having to forget the old ref.

Thanks for your patience, please lmk if there is anything else I can do. I think this should be everything I want to change in this PR.

src/entity_ref.rs
```diff
@@ -294,3 +349,79 @@ impl<'a, T: Component> ComponentRef<'a> for &'a mut T {
...
 impl<'a, T: Component> ComponentRefShared<'a> for &'a T {}
+
+struct ComponentBorrow<'a> {
+    archetype: &'a Archetype,
+    /// State index for `component` in `archetype`
+    state: usize,
+}
+
+impl<'a> ComponentBorrow<'a> {
+    unsafe fn for_component<T: Component>(
```
Ralith, line 360: A brief comment describing why this is unsafe (`index` is unchecked) wouldn't hurt

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# Unreleased
+
+### Added
+- `Ref::map` and `RefMut::map` to reborrow a subfield of a component, or cast it to an unsized type
+
```
Ralith, line 5: You could note the fix to `Ref::clone` here too.

Ralith (changes requested):
LGTM aside from some cosmetic tweaks, thanks! Good catch on that bug.

burtonageo:
Comments all (hopefully) addressed

src/entity_ref.rs
```diff
@@ -109,105 +110,159 @@ unsafe impl<'a> Send for EntityRef<'a> {}
...
-unsafe impl<T: Component> Send for Ref<'_, T> {}
-unsafe impl<T: Component> Sync for Ref<'_, T> {}
+unsafe impl<T: ?Sized + Sync> Send for Ref<'_, T> {}
+unsafe impl<T: ?Sized + Sync> Sync for Ref<'_, T> {}
 
-impl<'a, T: Component> Drop for Ref<'a, T> {
-    fn drop(&mut self) {
-        self.archetype.release::<T>(self.state);
+impl<'a, T: ?Sized> Ref<'a, T> {
+    /// Transform the `Ref<'_>` to point to to a part of the borrowed data, e.g.
```
Ralith, line 137: Typo:
```suggestion
    /// Transform the `Ref<'_>` to point to a part of the borrowed data, e.g.
```

src/entity_ref.rs
```diff
@@ -109,105 +110,159 @@ unsafe impl<'a> Send for EntityRef<'a> {}
...
-unsafe impl<T: Component> Send for RefMut<'_, T> {}
-unsafe impl<T: Component> Sync for RefMut<'_, T> {}
+unsafe impl<T: ?Sized + Send> Send for RefMut<'_, T> {}
+unsafe impl<T: ?Sized + Sync> Sync for RefMut<'_, T> {}
 
-impl<'a, T: Component> Drop for RefMut<'a, T> {
-    fn drop(&mut self) {
-        self.archetype.release_mut::<T>(self.state);
+impl<'a, T: ?Sized> RefMut<'a, T> {
+    /// Transform the `RefMut<'_>` to point to to a part of the borrowed data, e.g.
```
Ralith, line 218: As above

src/entity_ref.rs
```diff
@@ -294,3 +349,83 @@ impl<'a, T: Component> ComponentRef<'a> for &'a mut T {
...
+
+struct ComponentBorrow<'a> {
+    archetype: &'a Archetype,
+    /// State index for the borrowed component in the `archetype`.
+    state: usize,
+}
+
+impl<'a> ComponentBorrow<'a> {
+    // This method is unsafe as the `index` parameter is not validated
+    // to actually point to the correct component in the `archetype`.
```
Ralith, line 361: This is a bit confusing. No in-bounds value of `index` is "incorrect", and an out of bounds value does not address an incorrect *component*, but rather causes immediate undefined behavior. Bounds checking is what's being omitted here; we always have the right component type due to `Archetype::get_state::<T>`.

Ralith (changes requested):

src/entity_ref.rs
burtonageo: Addressed again!

I hope that the `{ComponentBorrow, ComponentBorrowMut}::for_component()` documentation is okay. I would have cribbed the documentation for `{Ref, RefMut}::new()` if there was any. If it's not okay, would you be able to suggest a comment I could copy paste for them? I'm not super familiar with how this part works or the safety guarantees.

Ralith (approved):
LGTM, thanks!

Ralith:
Published in 0.10.3.

PR #325, opened 2023-06-06
Update to edition 2021

a1phyr:
MSRV is already >1.56 anyway


Ralith (approved):
Thanks! Maybe get the `macros` crate as well?

a1phyr:
Done !

PR #329, opened 2023-07-02
Execute recorded commands in order

Ralith:


src/command_buffer.rs
```diff
@@ -134,67 +135,66 @@ impl CommandBuffer {
         unsafe {
             components.put(|ptr, ty| self.add_inner(ptr, ty));
         }
-        self.entities.push(EntityIndex {
+        self.components[first_component..].sort_unstable_by_key(|c| c.ty);
```
LPGhatguy, line 138: Are there performance implications for pushing this sort out of `run_on` and into `insert`/`spawn`? The total size of a `CommandBuffer` is likely small, but I wonder if this has some degenerate cases.

LPGhatguy (commented):
Nice! I think the code gets overall a little bit more readable with this change as a nice bonus.

src/command_buffer.rs
Ralith: It's been a long time since this code was written, but the only difference I can think of is if you drop or `clear` the `CommandBuffer` without using it, which really shouldn't take place on a hot path. In normal use, every range of components is sorted exactly once regardless.

LPGhatguy: Oh, I see how that works! That makes sense to me, cool.

LPGhatguy (approved):
LGTM!

PR #337, opened 2023-09-17
Implement queries that add/remove components

### LINKED ISSUE #334: Queries that add/remove components
Common patterns involve iterating over a query and adding, adding if absent, or removing a component from every entity. This presently requires a somewhat awkward and suboptimal dance where operations are buffered in a `CommandBuffer` or moral equivalent and executed in a separate pass after the query completes. We could extend the query machinery to support these patterns directly with query transformers that allow components to be directly written to or consumed from archetype storage, though considerable care will be required to soundly handle corner cases including incomplete iteration, leaked queries, and failure to insert all expected components.

Ralith:
Fixes #334.

TODO:
- [ ] Move side effects into a guard object exposed via a new API (`query_with_effect`?) to fix unsoundness
- [ ] Work out why this machinery is 10x slower in the `iterate_mut_100k` benchmark than `query_mut` in master. Nothing's jumping out at be in the disassembly; input would be welcome.

### DIFF
--- a/src/entities.rs
+++ b/src/entities.rs
@@ -548,8 +548,10 @@ impl EntityMeta {
     };
 }
 
+/// Addresses storage for an entity's components within a world
 #[derive(Copy, Clone)]
-pub(crate) struct Location {
+#[doc(hidden)]
+pub struct Location {
     pub archetype: u32,
     pub index: u32,
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,8 +89,8 @@ pub use entity_builder::{BuiltEntity, BuiltEntityClone, EntityBuilder, EntityBui
 pub use entity_ref::{ComponentRef, ComponentRefShared, EntityRef, Ref, RefMut};
 pub use query::{
     Access, Batch, BatchedIter, Or, PreparedQuery, PreparedQueryBorrow, PreparedQueryIter,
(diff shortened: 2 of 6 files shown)


adamreichold:
>  Work out why this machinery is 10x slower in the iterate_mut_100k benchmark than query_mut in master. Nothing's jumping out at me in the disassembly; input would be welcome.

This is pure speculation, but could this be cache effects of the increased working set due to the effect table?

Ralith:
> This is pure speculation, but could this be cache effects of the increased working set due to the effect table?

Surprisingly, that slowdown is for in-place/effect-free queries, for which effects are ZSTs and the effect table should allocate no memory.

a1phyr:
These benchmarks do very few things within the iteration, which sometimes allow LLVM to optimize the loop far better... or not !
This seems to be *very* fragile. Adding a `continue` in the iteration loop or a `black_box` with the query values can make or break it.

For example, on #341, adding `black_box` around `vel.0` in the benchmark code canceled the optimization.

Ralith:
A 10-fold difference is a lot to hand-wave, even so. Is that not sustained with `black_box` involved? Maybe a more computationally intensive benchmark would be interesting?

Ralith:
This would be useful for https://github.com/Ralith/hecs/pull/366, which presently must make two passes and perform a lot of redundant lookups to take ownership of components from entities satisfying a query.

PR #339, opened 2023-09-18
no-std support for macros

### LINKED ISSUE #106: no-std support for macros

wenyuzhao:
This PR fixes #106 and supports no_std builds to use macros.

Changes involved:
* Changed all `::std` to `::core` in macros crate
* Changed all `::std::vec` to `::hecs::alloc::vec`
* Replace `OnceLock` with `spin::Lazy`

~There is probably no need to deal with lazy_static as it already has the `spin_no_std` feature enabled. It should already support no_std builds.~ 

~After merging the commit to use `OnceLock` instead of `lazy_static`, I had to use `spin::Once` for no-std builds. `OnceLock` is only in `std`, not in `core`.~


macros/src/bundle.rs
```diff
@@ -79,23 +79,30 @@ fn gen_bundle_impl(
...
             ids
         }
     };
     let with_static_ids_body = if generics.params.is_empty() {
-        quote! {
-            static ELEMENTS: ::std::sync::OnceLock<[::std::any::TypeId; #num_tys]> = ::std::sync::OnceLock::new();
-            f(ELEMENTS.get_or_init(|| #with_static_ids_inner))
+        if cfg!(feature = "std") {
+            quote! {
+                static ELEMENTS: ::core::sync::OnceLock<[::core::any::TypeId; #num_tys]> = ::core::sync::OnceLock::new();
```
adamreichold, line 98: Does this path (starting with `::core`) really exist for `OnceLock`?

More generally, does it make sense to keep both code paths? The `spin`-based one should work fine on `std`-enabled targets so it might improve maintainability to just have that one code path. Furthermore, you might want to use `spin::lazy::Lazy` instead of `spin::once::Once` as it is more targeted towards this particular use case.

wenyuzhao: Sorry this is a typo... it should be `::std`.

I've removed it and used `spin::Laz for both std and no-std cases. The `hecs-macros/std` macro is also removed.

adamreichold (commented):
The changes look good to me. I wonder about the testing story though. Can we add a no_std build to the CI so that this does not regress so easily?

src/lib.rs
```diff
@@ -107,7 +108,7 @@ pub use bundle::DynamicClone;
 #[doc(hidden)]
 pub use query::Fetch;
 
-#[cfg(feature = "macros")]
+#[cfg(any(feature = "macros", feature = "macros_no_std"))]
```
adamreichold, line 111: Is the second feature flag here a left-over of some kind? I don't see where it is introduced.

adamreichold (approved):

Cargo.toml
```diff
@@ -32,15 +32,23 @@ row-serialize = ["serde"]
...
 serde = { version = "1.0.117", features = ["derive"] }
 serde_test = "1.0.117"
 
+[target.'cfg(target_os = "none")'.dev-dependencies]
+
+# An example to demonstrate that the macros crate can be used without std.
+# This example is mainly used for testing.
+[[example]]
+name = "no_std_macros"
+crate-type = ["staticlib"]
```
adamreichold, line 50: I think you need to add `required-features = ["macros"]` to avoid build failures in `no-default-features` builds.

src/lib.rs
```diff
@@ -39,7 +39,8 @@
 #[cfg(feature = "std")]
 extern crate std;
 
-extern crate alloc;
+pub extern crate alloc;
+pub extern crate spin;
```
Ralith, line 43: Let's not commit to these as part of the public API.
```suggestion
#[doc(hidden)]
pub extern crate alloc;
#[doc(hidden)]
pub extern crate spin;
```

examples/no_std_macros.rs
Ralith, line 1: Are all of these three bundles contributing to coverage, or could some be stripped down?

examples/no_std_macros.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub bar: &'a mut bool,
+}
+
+#[panic_handler]
+#[cfg(target_os = "none")]
+fn panic(_: &core::panic::PanicInfo) -> ! {
+    loop {}
+}
+
+struct NoAlloc;
```
Ralith, line 33: Could we have `extern crate alloc;` rather than this stubbed machinery? hecs won't work without dynamic allocation, so this might cause some confusion.

examples/no_std_macros.rs
Ralith, line 1: This file should have a module-level comment explaining that it's for testing purposes, rather than being intended to teach something to users.

.github/workflows/rust.yml
```diff
@@ -35,6 +35,30 @@ jobs:
...
+          target: x86_64-unknown-none
+          override: true
+
+      # Build the `no_std_macros` example with x86_64-unknown-none target (should not contain std)
+      - uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --no-default-features --features macros --target x86_64-unknown-none --example no_std_macros
+
+      # Build the same `no_std_macros` example with the default target (which should contain std)
```
Ralith, line 56: Is this already covered by the generic test job?

Ralith (commented):
Could an integration test be used instead of an example? If not, we should at least add a comment to explain that this example isn't really demonstrating much. Another approach would be to have a dedicated `publish = false` crate on the side.

src/lib.rs
wenyuzhao: Yes this is a left-over. Removed.

examples/no_std_macros.rs
wenyuzhao: Removed.

This is necessary for a `staticlib` build. `extern crate alloc;` does not solve the problem. But looks like a `rlib` build is already enough. Any use of the `::std` in the macro expansion will fail the no_std rlib build.

examples/no_std_macros.rs
wenyuzhao: This was directly copied from one of the tests. One `#[derive(Bundle)]` should already be enough. I removed the other two.

src/lib.rs
wenyuzhao: Fixed.

examples/no_std_macros.rs
wenyuzhao: I managed to move this to a `publish = false` + `no_std` crate, and added an integration test to build the crate. This example file is removed now.

.github/workflows/rust.yml
wenyuzhao: Yes looks like this is not necessary.

I removed all changes to the workflow config file, as the `no_std` build test is a part of the integration test now.

tests/no_std_builds.rs
```diff
@@ -0,0 +1,11 @@
+use std::process::Command;
+
+#[test]
+fn build_no_std_macros() {
+    // Ensure that the macro crate and the macro-expanded code can work in a `no_std` environment.
+    let status = Command::new("cargo")
+        .args(["build", "-p", "test-no-std-macros"])
```
wenyuzhao, line 7: FYI looks like building the default `x86_64-unknown-linux-gnu` target is enough here. The `test-no-std-macros` crate is marked as `#[no_std]`. Any use of `::std` in the macro-expanded code should fail the compilation, even with the `x86_64-unknown-linux-gnu` target. There is no need to build with the `x86_64-unknown-none` target here.

Cargo.toml
wenyuzhao: This example is removed. It's now part of the integration test.

tests/no_std_builds.rs
Ralith: Running cargo from inside the build script is very obscure and will fail in some environments. Let's add a CI job instead.

Ralith (changes requested):

tests/no_std_builds.rs
wenyuzhao: Sure. The build test is moved to CI.

Ralith (approved):

PR #342, opened 2023-09-21
Add block_box to benches

a1phyr:
This makes `iterate_*` benches far more stable and less code layout dependent.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -7,6 +7,7 @@
 
 use bencher::{benchmark_group, benchmark_main, Bencher};
 use hecs::*;
+use std::hint::black_box;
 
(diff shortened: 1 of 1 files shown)


Ralith:
I'm not sure this the right place for it. We *want* the compiler to be able to aggressively optimize the interior of the loop; if it can't, that means the iterator code is probably too complex and obscuring things. In the benchmarks we should only be concerned about work being deleted entirely, rather than just rearranged, and I haven't seen evidence that. If we did see that it could probably be handled with a single `black_box` application to the entire world after the loop.

Separately, maybe we should use the now stable and presumably more reliable `std::hint` version?

PR #344, opened 2023-09-25
Improve CI

a1phyr:
This fixes the clippy error in CI, uses last version of actions and removes/replaces deprecated ones

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,88 +17,57 @@ jobs:
           - --no-default-features
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v4
 
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@stable
         with:
-          profile: minimal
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -17,88 +17,57 @@ jobs:
...
         with:
-          profile: minimal
           toolchain: ${{ matrix.rust }}
-          override: true
 
-      - uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: -p hecs --all-targets ${{ matrix.features }}
+      - run: rustup install --profile minimal ${{ matrix.rust }} && rustup default ${{ matrix.rust }}
```
adamreichold, line 26: Why not do this using the parameters of `dtolnay/rust-toolchain`?

a1phyr: I started by calling `rustup` directly and ended up using `dtolnay/rust-toolchain`, this is a leftover that I forgot

Ralith (approved):
Thanks!

PR #346, opened 2023-10-10
Use `AtomicIsize::get_mut` where appropriate

rj00a:
I noticed a few atomic loads and stores could be eliminated in `entities.rs` because mutable access to `free_cursor` is available.


Ralith (approved):
Nice catch, thanks!

PR #354, opened 2023-12-03
foo_bundle_satisfies_query added

sanbox-irl:
Added two new bare functions, accessed as `hecs:: bundle_satisfies_query` and `hecs::dynamic_ bundle_satisfies_query`. This should help in certain unit tests

### DIFF
--- a/src/bundle.rs
+++ b/src/bundle.rs
@@ -14,30 +14,6 @@ use crate::archetype::TypeInfo;
 use crate::Component;
 
 /// Checks if a query is satisfied by a bundle. This is primarily useful for unit tests.
-///
-/// ```rs
-/// #[derive(hecs::Bundle)]
-/// struct MyBundle {
-///     health: i32,
-///     speed: f32,
-/// }
-///
-/// #[derive(hecs::Query)]
-/// struct MyQuery<'a> {
-///     health: &'a i32,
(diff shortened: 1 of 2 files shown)


src/bundle.rs
```diff
@@ -13,6 +13,75 @@ use core::{fmt, mem};
...
+///     health: &'a i32,
+///     speed: &'a f32,
+/// }
+///
+/// #[cfg(test)]
+/// mod tests {
+///     use super::{MyBundle, MyQuery};
+///     
+///     #[test]
+///     fn my_bundle_satisfies_my_query() {
```
Ralith, line 36: For concision, I think we can omit all this boilerplate. Just the bare assert after the definitions should be fine.

src/bundle.rs
```diff
@@ -13,6 +13,75 @@ use core::{fmt, mem};
 use crate::archetype::TypeInfo;
 use crate::Component;
 
+/// Checks if a query is satisfied by a bundle. This is primarily useful for unit tests.
+///
+/// ```rs
+/// #[derive(hecs::Bundle)]
```
Ralith, line 19: Won't using these macros in doctests cause `cargo test` to fail when the macros aren't enabled? Maybe we should use type aliases instead.

PR #356, opened 2023-12-06
clippy yells no more

sanbox-irl:
very minor change, but makes working in the repo nicer if you have clippy auto-running since there aren't 2 warnings always


Ralith (approved):
Thanks!

Issue #317, opened 2023-04-18
Implementation of `Ord` on `Entity`


LPGhatguy:
Currently, the `Ord` implementation on `Entity` behaves the same way as the `to_bits` representation of the entity: sorted by generation first, then by ID. That parity is nice, but I am unsure if this is the right default order.

I think it might make sense to sort entities by their ID first and then their generation. This makes the behavior of data structures like `BTreeMap` with entity keys behave more deterministically.

In our game, we have a `WireNetwork` component that holds a `BTreeMap<EntityHandle, Connections>`, where `EntityHandle` is a new-typed wrapper around `Entity` with custom serialization behavior. We were deriving `Ord` on that type, but I was surprised when the entries in the map would get shuffled around. This ended up being due to generations sometimes being different between runs, reordering the map!

We were able to solve that issue by manually implementing `Ord` to compare the entity's ID first. Doing that made me wonder whether  the same change would be right for hecs itself.

Ralith:
That sounds like a reasonable change to me! I wonder if we should flip the bit representation around to match?

Ralith:
Fixed by #357.

PR #357, opened 2023-12-06
flipped entities repr for ordering

sanbox-irl:
needed, fulfills #317.

fwiw -- i feel *pretty* iffy about changing the bit repr. I know that hecs says that it "doesn't promise anything" but this library is used a lot! personally, i'd leave the old bit repr in place


sanbox-irl:
If you're comfortable with that, I'm comfortable with that, but I'd say that the Ord change is "higher level" than the bit fiddling changes, so it feels simpler to understand.

sanbox-irl:
@Ralith bumpin this for your thoughts

Ralith (approved):

PR #361, opened 2023-12-23
query-n-mut

sanbox-irl:
This is tracking my work. I've done an initial implementation but I'm not very familiar with this code, so I'm not even sure if this is right or legal. Additionally, there ~probably~ should be a test in here, somewhere.


sanbox-irl:
closing in favor of #362

PR #364, opened 2024-01-20
Determinism for Serialization

ZagButNoZig:
This a draft for making the serialization of a world deterministic by adding the necessary information to the serialization.
It works by saving the free list and all the generations of all alive and de spawned entities.

### TODOs

- [x] Fix the 2 serialize test cases
- [x] Remove column serialize from default features
- [x] Make sure everything works as intended
- [x] ~~Maybe add this to row serialization aswell~~ (Out of scope for this PR)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,6 @@ spin = { version = "0.9.8", default-features = false, features = ["mutex", "spin
 bencher = "0.1.5"
 rand = "0.8.3"
 trybuild = "1.0.23"
-serde_json = "1.0"
 serde = { version = "1.0.117", features = ["derive"] }
 serde_test = "1.0.117"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,13 +31,14 @@ row-serialize = ["serde"]
 [dependencies]
 hecs-macros = { path = "macros", version = "0.10.0", optional = true }
 hashbrown = { version = "0.14", default-features = false, features = ["ahash", "inline-more"] }
-serde = { version = "1.0.117", default-features = false, optional = true }
+serde = { version = "1.0.117", default-features = false, features = ["alloc"], optional = true }
 spin = { version = "0.9.8", default-features = false, features = ["mutex", "spin_mutex", "lazy"] }
 
 [dev-dependencies]
 bencher = "0.1.5"
 rand = "0.8.3"
 trybuild = "1.0.23"
+serde_json = "1.0"
 serde = { version = "1.0.117", features = ["derive"] }
 serde_test = "1.0.117"
 
--- a/src/entities.rs
+++ b/src/entities.rs
@@ -109,6 +109,12 @@ impl<'de> serde::Deserialize<'de> for Entity {
     }
 }
 
+#[derive(Debug)]
+pub enum PendingPushError {
+    MetaTooSmall,
+    EntityNotPending
+}
+
 /// An iterator returning a sequence of Entity values from `Entities::reserve_entities`.
 pub struct ReserveEntitiesIterator<'a> {
     // Metas, so we can recover the current generation for anything in the freelist.
@@ -146,6 +152,12 @@ impl<'a> Iterator for ReserveEntitiesIterator<'a> {
 }
 
 impl<'a> ExactSizeIterator for ReserveEntitiesIterator<'a> {}
+
+
+pub struct Freelist<'a> {
+    pub pending: &'a [u32],
+    pub free_cursor: isize,
+}
 
 #[derive(Default)]
 pub(crate) struct Entities {
@@ -192,6 +204,13 @@ pub(crate) struct Entities {
 }
 
 impl Entities {
+    pub fn freelist(&self) -> Freelist {
+        Freelist {
+            pending: &self.pending,
+            free_cursor: self.free_cursor.load(Ordering::Relaxed),
+        }
+    }
+
     /// Reserve entity IDs concurrently
     ///
     /// Storage for entity generation and location is lazily allocated by calling `flush`.
@@ -343,10 +362,7 @@ impl Entities {
         self.verify_flushed();
 
         let loc = if entity.id as usize >= self.meta.len() {
-            self.pending.extend((self.meta.len() as u32)..entity.id);
-            let new_free_cursor = self.pending.len() as isize;
-            *self.free_cursor.get_mut() = new_free_cursor;
-            self.meta.resize(entity.id as usize + 1, EntityMeta::EMPTY);
+            self.extend_meta((entity.id + 1) as usize, false);
             self.len += 1;
             None
         } else if let Some(index) = self.pending.iter().position(|item| *item == entity.id) {
@@ -494,7 +510,7 @@ impl Entities {
         }
     }
 
-    fn needs_flush(&mut self) -> bool {
+    pub fn needs_flush(&mut self) -> bool {
         *self.free_cursor.get_mut() != self.pending.len() as isize
     }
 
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -19,7 +19,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["std"]
+default = ["std", "column-serialize"]
```
Ralith, line 22: Please do not add to the default features.

src/entities.rs
```diff
@@ -182,7 +188,7 @@ pub(crate) struct Entities {
     // and then from the new IDs, using only a single atomic subtract.
     //
     // Once `flush()` is done, `free_cursor` will equal `pending.len()`.
-    pending: Vec<u32>,
+    pub pending: Vec<u32>,
```
Ralith, line 191: Exposing this directly vastly complicates reasoning about the data structure's invariants. Instead, add methods that expose the required functionality without interfering with invariants.

src/entities.rs
```diff
@@ -490,8 +493,8 @@ impl Entities {
         }
     }
 
-    fn needs_flush(&mut self) -> bool {
-        *self.free_cursor.get_mut() != self.pending.len() as isize
+    pub fn needs_flush(&self) -> bool {
```
Ralith, line 496: This judgement isn't meaningful if the caller doesn't have unique access.

src/entities.rs
```diff
@@ -521,6 +524,49 @@ impl Entities {
...
+        self.meta.resize(new_len, EntityMeta::EMPTY);
+    }
+
+    pub fn push_generations(&mut self, generations: &[u32]) {
+        if self.meta.len() < generations.len() {
+            self.extend_meta(generations.len(), true);
+        }
+
+        for (index, meta) in self.meta.iter_mut().enumerate() {
+            meta.generation = NonZeroU32::new(generations[index]).expect("Could not create non zero u32");
```
Ralith, line 546: This panic message is not helpful. Either write a message that gives a developer seeing it useful context, or use a concise `unwrap`. Panic messages should contain information that isn't already immediately obvious from the backtrace.

src/serialize/column.rs
```diff
@@ -291,14 +297,21 @@ where
     }
 
     let predicate = |x: &&Archetype| -> bool { !x.is_empty() && x.satisfies::<Q>() };
-    let mut seq = serializer.serialize_seq(Some(world.archetypes().filter(predicate).count()))?;
+    let mut seq = serializer.serialize_seq(Some(world.archetypes().filter(predicate).count() + 2))?;
+    
+    let generations: Vec<u32> = world.entities_meta().iter().map(|m| m.generation.get()).collect();
```
Ralith, line 302: Serialization logic must not use non-public API. Anything these modules do must be achievable by external code, because users may need to serialize a world using tools other than serde.

src/serialize/column.rs
```diff
@@ -291,14 +297,21 @@ where
     }
 
     let predicate = |x: &&Archetype| -> bool { !x.is_empty() && x.satisfies::<Q>() };
-    let mut seq = serializer.serialize_seq(Some(world.archetypes().filter(predicate).count()))?;
+    let mut seq = serializer.serialize_seq(Some(world.archetypes().filter(predicate).count() + 2))?;
```
Ralith, line 300: Changes to the serialization schema must be reflected in the module level documentation.

src/serialize/column.rs
```diff
@@ -761,6 +783,7 @@ mod tests {
     enum ComponentId {
         Position,
         Velocity,
+        I32
```
Ralith, line 786: What purpose does adding an additional component type to these tests serve?

src/serialize/column.rs
```diff
@@ -911,10 +941,41 @@ mod tests {
...
 
+    #[test]
+    fn deterministic() {
+        let buffer = Vec::<u8>::new();
+
+        let mut world = World::new();
+        let mut context = Context {
+            components: vec![ComponentId::Position, ComponentId::Velocity]
+        };
+        let mut serializer = serde_json::Serializer::pretty(buffer);
```
Ralith, line 957: This is a large increase in dev-dependencies. Why can't we use `serde_test` here like every other test?

src/world.rs
```diff
@@ -413,10 +413,25 @@ impl World {
     }
 
     #[inline(always)]
-    pub(crate) fn entities_meta(&self) -> &[EntityMeta] {
+    pub(crate) fn entities_meta(&self) -> &Vec<EntityMeta> {
         &self.entities.meta
     }
 
+    /// Get the id's of all pending (free) entities
+    pub fn pending(&self) -> &Vec<u32> {
```
Ralith, line 421: `&Vec<u32>` is almost never a useful type in Rust. Prefer `&[u32]`.

Ralith, line 421: `World::pending` is a deeply obscure name. How about `World::entity_freelist`? Regardless, the doc comment should explain what this is and why it's useful.

src/world.rs
```diff
@@ -413,10 +413,25 @@ impl World {
...
+
+    /// Set a custom list of generations
+    pub fn push_generations(&mut self, generations: &[u32]) {
+        self.entities.push_generations(generations);
+    }
+
+    /// Manually set what entities are pending
+    pub fn push_pending(&mut self, pendings: &[u32]) -> Result<(), PendingPushError>{
+        self.entities.push_pending(pendings)
+    }
```
Ralith, line 433: Is there a use case for calling these on a pre-existing world, or multiple times, or calling one but not the other? If not, these should be arguments to an alternative constructor, not methods.

src/serialize/column.rs
```diff
@@ -186,6 +188,10 @@ where
     S: Serializer,
     C: SerializeContext,
 {
+    if !world.is_flushed() {
+        return Result::Err(<S as Serializer>::Error::custom("Serialize requires the world to be flushed first."));
```
Ralith, line 192: It's not obvious why this is necessary, or whether it's even necessary at all. Recall that most users don't need serialization to preserve entity allocator state to begin with.

Cargo.toml
```diff
@@ -31,13 +31,14 @@ row-serialize = ["serde"]
...
 hashbrown = { version = "0.14", default-features = false, features = ["ahash", "inline-more"] }
-serde = { version = "1.0.117", default-features = false, optional = true }
+serde = { version = "1.0.117", default-features = false, features = ["alloc"], optional = true }
 spin = { version = "0.9.8", default-features = false, features = ["mutex", "spin_mutex", "lazy"] }
 
 [dev-dependencies]
 bencher = "0.1.5"
 rand = "0.8.3"
 trybuild = "1.0.23"
+serde_json = "1.0"
```
Ralith, line 41: This should not be necessary.

Ralith (changes requested):

src/entities.rs
ZagButNoZig: I also serialize the free_cursor now, so this is not necessary anymore.

src/serialize/column.rs
ZagButNoZig: Fixed.

src/serialize/column.rs
ZagButNoZig: Removed.

src/world.rs
ZagButNoZig: Replaced by freelist.

src/world.rs
ZagButNoZig: Moved both of them into one method.

src/serialize/column.rs
ZagButNoZig: Removed.

src/entities.rs
ZagButNoZig: Replaced with unwrap.

src/world.rs
ZagButNoZig: Renamed and tried to improve the doc of Freelist.

src/serialize/column.rs
ZagButNoZig: Updated.

caspark:
@Ralith is this PR sound/ready to be merged? I would benefit from having this in master too, so if it's falling short in some way then I can try to fix it up.

Ralith:
@ZagButNoZig, thanks for your work exploring this problem space! Given the age of this PR, rather than requesting further iteration I've drafted https://github.com/Ralith/hecs/pull/390 for the tricky parts.

@caspark Thanks for your interest! https://github.com/Ralith/hecs/pull/390 should unlock the basic capability; please feel free to weigh in there if you have any thoughts.

PR #366, opened 2024-02-04
Introduce change tracking helper

Ralith:
Much like `CommandBuffer` and serialization support, I think this is non-obvious, useful, and self-contained enough to merit inclusion even though it's implemented purely in terms of the public API. I'm interested in feedback on whether this would indeed be useful to people, and on the API design, including the detailed semantics.

### DIFF
--- a/src/change_tracker.rs
+++ b/src/change_tracker.rs
@@ -1,0 +1,212 @@
+use core::mem;
+
+use alloc::vec::Vec;
+
+use crate::{Component, Entity, PreparedQuery, With, Without, World};
+
+/// Helper to track changes in `T` components
+///
+/// For each entity with a `T` component, a private component is inserted which stores the value as
+/// of the most recent call to `track`. This provides robust, exact change detection at the cost of
+/// visiting each possibly-changed entity. It is a good fit for entities that will typically be
+/// visited regardless, and components having fast [`Clone`] and [`PartialEq`] impls. For components
+/// which are expensive to compare and/or clone, consider instead tracking changes manually, e.g.
+/// by setting a flag in the component's `DerefMut` implementation.
+///
+/// Always use exactly one `ChangeTracker` per [`World`] per component type of interest. Using
+/// multiple trackers of the same `T` on the same world, or using the same tracker across multiple
+/// worlds, will produce unpredictable results.
+pub struct ChangeTracker<T: Component> {
+    added: PreparedQuery<Without<&'static T, &'static Previous<T>>>,
+    changed: PreparedQuery<(&'static T, &'static mut Previous<T>)>,
+    removed: PreparedQuery<Without<With<(), &'static Previous<T>>, &'static T>>,
+
+    added_components: Vec<(Entity, T)>,
+    removed_components: Vec<Entity>,
+}
+
+impl<T: Component> ChangeTracker<T> {
+    /// Create a change tracker for `T` components
+    pub fn new() -> Self {
+        Self {
+            added: PreparedQuery::new(),
+            changed: PreparedQuery::new(),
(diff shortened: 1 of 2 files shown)


kanerogers:
My 2 cents

What I like:
- It's opt-in
- Self-contained
- Documentation on `track` is excellent
- Conceptually simple

What I don't like:
- Some module-level documentation would help with visibility (happy to volunteer for this)

Ralith:
> module-level documentation

The module isn't public; I'm just reexporting `ChangeTracker` and `Changes` at the top level. I'd welcome a succinct doctest for `ChangeTracker` to illustrate its use, though.

src/change_tracker.rs
```diff
@@ -0,0 +1,209 @@
...
+    ///
+    /// For each entity with a `T` component, a private component is inserted which stores the value
+    /// as of the most recent call to `track`. This provides robust, exact change detection at the
+    /// cost of visiting each possibly-changed entity. It is a good fit for entities that will
+    /// typically be visited regardless, and `T`s having fast [`Clone`] and [`PartialEq`] impls.
+    ///
+    /// Always use exactly one `ChangeTracker` per [`World`] per component type of interest. Using
+    /// multiple trackers of the same `T` on the same world, or using the same tracker across
+    /// multiple worlds, will produce unpredictable results.
+    pub fn track<'a>(&'a mut self, world: &'a mut World) -> Changes<'a, T>
```
adamreichold, line 40: Is the exclusive reference to the `World` required for correctness? Wouldn't a shared reference suffice but allow interleaving the tracking of changes to multiple types `T` on a given `World`?

Ralith: The exclusive reference is required to insert/remove `Previous<T>`s as needed for the added/removed cases. Coupling those updates tightly with the queries makes it very difficult to inadvertently double-count a change.

src/change_tracker.rs
```diff
@@ -0,0 +1,209 @@
...
+impl<'a, T: Component> Drop for Changes<'a, T>
+where
+    T: Component + Clone + PartialEq,
+{
+    fn drop(&mut self) {
+        if !self.added {
+            _ = self.added();
+        }
+        for (entity, component) in self.tracker.added_components.drain(..) {
+            self.world.insert_one(entity, Previous(component)).unwrap();
```
adamreichold, line 135: Not strictly part of this change but something that might be motivated by it: If not queries that add/remove components, maybe a bulk insertion/removal method might be helpful here. I was thinking about sorting the entities by archetype and then handling all entities in a given archetype as a single operation? (This would probably require quite a few extra code paths, so I am not sure about the cost/benefit ratio.)

Ralith: > maybe a bulk insertion/removal method might be helpful here

Yeah, I want this. Repeatedly inserting/removing from the world scope is definitely suboptimal. See also the TODO, and https://github.com/Ralith/hecs/pull/337. A narrower solution than https://github.com/Ralith/hecs/pull/337 might be sufficient to get a nice fast-path here, but I'm not sure how to come up with a good API design for bulk insert where a new value is dependent on existing components. Maybe require a constructor closure, and let the query overwrite it? Bulk removal is easier API-design-wise but probably a nontrivial chunk of code all the same.

h3r2tic:
Looks quite handy, and I like how small and clear an implementation this is! FWIW our project enjoys change tracking in Bevy via `DerefMut` marking components dirty (as opposed to running `PartialEq`). It most cases it wouldn't matter much, but at least in one one case the component involved is a chonky one, and I suspect that with this implementation, we'd need to layer an additional dirtiness tracking thing on top in order to avoid the cost of `PartialEq`.

Issue #328, opened 2023-07-02
Ability to create a query view directly from `World`


LPGhatguy:
hecs has a great mechanism for random access of a query via views:

```rust
let mut query = game.world.query::<(&A, &B, &C)>();
let view = query.view();
let (a, b, c) = view.get(some_entity)?;
```

It'd be really handy to collapse the calls to `query` and `view` into just one call, to avoid needing to create the extra binding.

I think there are a couple good solutions to this problem:
- A new method, `World::view` and a new `View` type that owns its contents
- Refactoring of `QueryBorrow` to enable random access

Ralith:
> Refactoring of QueryBorrow to enable random access

This would involve taking the `fetch` field of `View`, which is a table with one entry per archetype, and adding it to `query`, populating it either up front (which adds unnecessary work to normal queries) or lazily (which adds a branch to every random access, the case that views are specifically supposed to optimize). Given those (minor) drawbacks, I think I lean slightly towards paying the modest one-time cost of introducing a new type (perhaps `ViewBorrow`).

Are you interested in making a PR for this?

LPGhatguy:
I'd be interested in working on this feature, but I'm pretty booked for a couple months. If no one else comes along and takes it by then, I can take a look!

LPGhatguy:
I created a cursed and unsafe version of this API in the hecs wrapper we have in our engine. We use views a lot, so I think it will be very nice to have.

I'm going to use it for awhile and see whether it's a very large gain. If it is, I'll take a go at implementing it properly in hecs!

Ralith:
This would incrementally tidy up some of the code I've been writing lately too.

Ralith:
Fixed by #367.

PR #367, opened 2024-03-22
view on world

sanbox-irl:
solves #328

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -1713,37 +1713,6 @@ impl<'w, Q: Query> core::ops::DerefMut for ViewBorrowed<'w, Q> {
     }
 }
 
-/// A borrow of a [`World`](crate::World) sufficient to execute the query `Q`, and immediately exposes a `View`
-/// of it.
-///
-/// This struct is a thin wrapper around [`View`](crate::View). See it for more documentation.
-pub struct ViewMut<'w, Q: Query> {
-    view: View<'w, Q>,
-}
-
-impl<'w, Q: Query> ViewMut<'w, Q> {
-    pub(crate) fn new(world: &'w mut World) -> Self {
-        assert_borrow::<Q>();
-        let view = unsafe { View::<Q>::new(world.entities_meta(), world.archetypes_inner()) };
-
-        Self { view }
-    }
-}
-
-impl<'w, Q: Query> core::ops::Deref for ViewMut<'w, Q> {
-    type Target = View<'w, Q>;
-
-    fn deref(&self) -> &Self::Target {
-        &self.view
-    }
-}
-
-impl<'w, Q: Query> core::ops::DerefMut for ViewMut<'w, Q> {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.view
-    }
-}
-
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/src/query.rs
+++ b/src/query.rs
@@ -1645,7 +1645,7 @@ impl<'a, 'q, Q: Query> IntoIterator for &'a mut PreparedView<'q, Q> {
 ///
 /// Note that borrows are not released until this object is dropped.
 ///
-/// This struct is a thin wrapper around [`View`](crate::View). See it for more documentation.
+/// This struct is a thin wrapper around [`View`]. See it for more documentation.
 pub struct ViewBorrow<'w, Q: Query> {
     view: View<'w, Q>,
 }
@@ -1664,7 +1664,7 @@ impl<'w, Q: Query> ViewBorrow<'w, Q> {
     ///
     /// Does not require exclusive access to the map, but is defined only for queries yielding only shared references.
     ///
-    /// See [View::get].
+    /// See [`View::get``].
     pub fn get(&self, entity: Entity) -> Option<Q::Item<'_>>
     where
         Q: QueryShared,
@@ -1676,21 +1676,21 @@ impl<'w, Q: Query> ViewBorrow<'w, Q> {
     ///
     /// Will yield `None` if the entity does not exist or does not match the query.
     ///
-    /// See [View::get_mut].
+    /// See [`View::get_mut``].
     pub fn get_mut(&mut self, entity: Entity) -> Option<Q::Item<'_>> {
         self.view.get_mut(entity)
     }
 
     /// Equivalent to `get(entity).is_some()`, but does not require `Q: QueryShared`
     ///
-    /// See [View::contains].
+    /// See [`View::contains``].
     pub fn contains(&self, entity: Entity) -> bool {
         self.view.contains(entity)
     }
 
     /// Like `get_mut`, but allows simultaneous access to multiple entities
     ///
-    /// See [View::get_unchecked].
+    /// See [`View::get_unchecked``].
     ///
     /// # Safety
     ///
@@ -1703,7 +1703,7 @@ impl<'w, Q: Query> ViewBorrow<'w, Q> {
     ///
     /// For N > 3, the check for distinct entities will clone the array and take O(N log N) time.
     ///
-    /// See [View::get_many_mut].
+    /// See [`View::get_many_mut``].
     ///
     /// # Examples
     ///
@@ -1731,7 +1731,7 @@ impl<'w, Q: Query> ViewBorrow<'w, Q> {
 
     /// Iterate over all entities satisfying `Q`
     ///
-    /// See [View::iter_mut]
+    /// See [`View::iter_mut``]
     pub fn iter_mut(&mut self) -> ViewIter<'_, Q> {
         self.view.iter_mut()
     }
--- a/src/query.rs
+++ b/src/query.rs
@@ -649,15 +649,7 @@ impl<'w, Q: Query> QueryBorrow<'w, Q> {
         if self.borrowed {
             return;
         }
-        for x in self.world.archetypes() {
-            if x.is_empty() {
-                continue;
-            }
-            // TODO: Release prior borrows on failure?
-            if let Some(state) = Q::Fetch::prepare(x) {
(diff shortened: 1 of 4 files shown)


src/query.rs
```diff
@@ -1655,6 +1655,45 @@ fn assert_distinct<const N: usize>(entities: &[Entity; N]) {
...
     }
 }
+
+/// A borrow of a [`World`](crate::World) sufficient to execute the query `Q`, and immediately exposes a `View`
+/// of it.
+///
+/// Note that borrows are not released until this object is dropped.
+///
+/// This struct is a thin wrapper around [`View`](crate::View). See it for more documentation.
+pub struct ViewBorrowed<'w, Q: Query> {
```
Ralith, line 1665: If the intention is to match `QueryBorrow`, then this should be `ViewBorrow`.

src/query.rs
```diff
@@ -1655,6 +1655,45 @@ fn assert_distinct<const N: usize>(entities: &[Entity; N]) {
         }
     }
 }
+
+/// A borrow of a [`World`](crate::World) sufficient to execute the query `Q`, and immediately exposes a `View`
+/// of it.
```
Ralith, line 1660:
```suggestion
/// A borrow of a [`World`](crate::World) sufficient to random-access the results of the query `Q`
```

src/query.rs
```diff
@@ -1655,6 +1655,45 @@ fn assert_distinct<const N: usize>(entities: &[Entity; N]) {
...
+impl<'w, Q: Query> core::ops::Deref for ViewBorrowed<'w, Q> {
+    type Target = View<'w, Q>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.view
+    }
+}
+
+impl<'w, Q: Query> core::ops::DerefMut for ViewBorrowed<'w, Q> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
```
Ralith, line 1692: I think this may be unsound after all. A caller could create a `View<'w, Q>` from an unrelated world, then `mem::swap` it with this one, then drop the `ViewBorrow`, leading to dangling references, violating hecs' preconditions. That procedure could then be repeated to create illegally aliasing references, creating immediate UB.

The only way I see to avoid this is to ensure that `View` itself borrows from the `ViewBorrow`, not the `World`. I don't think that's expressible through `DerefMut`, and with an explicit getter we're back at exactly the same ergonomics as creating a `query` first.

The only sound path forwards I see is to manually replicate the `View` API directly on `ViewBorrow`. We could get up to some trait shenanigans, but I'm not a fan of requiring users to `use` a trait to call routine methods on a concrete type. This is mildly annoying to do, but IMO library boilerplate is a low cost to pay for a better user experience, and it's a tradeoff hecs already makes in various other places (e.g. prepared queries).

src/world.rs
```diff
@@ -399,6 +399,17 @@ impl World {
         QueryBorrow::new(self)
     }
 
+    /// Provide random access to any entity for a given Query.
+    pub fn view<Q: Query>(&self) -> ViewBorrowed<'_, Q> {
+        ViewBorrowed::new(self)
+    }
+    /// Provide random access to any entity for a given Query on a uniquely
```
Ralith, line 406: nit:
```suggestion
    }

    /// Provide random access to any entity for a given Query on a uniquely
```

src/world.rs
```diff
@@ -1360,4 +1371,68 @@ mod tests {
...
+        // edit some value
+        *int_view.get_mut(e0).unwrap() = 100;
+
+        let mut int_str_view = world.view_mut::<(&&str, &mut i32)>();
+        let (s, i) = int_str_view.get_mut(e0).unwrap();
+        assert_eq!(*s, "hello");
+        assert_eq!(*i, 100);
+        assert_eq!(int_str_view.get_mut(e1), None);
+        assert_eq!(int_str_view.get_mut(e2), None);
+    }
```
Ralith, line 1437: nit: the above are integration tests, not really particular to the internals of `World` specifically, and might hence make more sense in tests/tests.rs.

Ralith (changes requested):

src/query.rs
```diff
@@ -1637,6 +1622,106 @@ impl<'a, 'q, Q: Query> IntoIterator for &'a mut PreparedView<'q, Q> {
...
+    /// ```
+    /// # use hecs::World;
+    /// let mut world = World::new();
+    ///
+    /// let a = world.spawn((1, 1.0));
+    /// let b = world.spawn((2, 4.0));
+    /// let c = world.spawn((3, 9.0));
+    ///
+    /// let mut query = world.query_mut::<&mut i32>();
+    /// let mut view = query.view();
```
Ralith, line 1709: This example is inaccurate. Maybe the doc comments here should all just be "See [\`View::whatever\`]"? Less risk of drift that way, too.

tests/tests.rs
Ralith, line 1: Include a `#[should_panic]` test ensuring that we can't create aliasing views with this API.

Ralith (changes requested):
Thanks, this looks great! Just a couple nits.

src/query.rs
sanbox-irl: Sounds good

Ralith: Low cost to fixing if this does bite us later, anyway.

src/query.rs
```diff
@@ -1656,6 +1641,118 @@ impl<'a, 'q, Q: Query> IntoIterator for &'a mut PreparedView<'q, Q> {
...
+        Self { view }
+    }
+
+    /// Retrieve the query results corresponding to `entity`
+    ///
+    /// Will yield `None` if the entity does not exist or does not match the query.
+    ///
+    /// Does not require exclusive access to the map, but is defined only for queries yielding only shared references.
+    ///
+    /// See [View::get].
```
Ralith, line 1667:
```suggestion
    /// See [`View::get`].
```

src/query.rs
```diff
@@ -1656,6 +1641,118 @@ impl<'a, 'q, Q: Query> IntoIterator for &'a mut PreparedView<'q, Q> {
...
+        Q: QueryShared,
+    {
+        self.view.get(entity)
+    }
+
+    /// Retrieve the query results corresponding to `entity`
+    ///
+    /// Will yield `None` if the entity does not exist or does not match the query.
+    ///
+    /// See [View::get_mut].
```
Ralith, line 1679:
```suggestion
    /// See [`View::get_mut`].
```

src/query.rs
```diff
@@ -1656,6 +1641,118 @@ impl<'a, 'q, Q: Query> IntoIterator for &'a mut PreparedView<'q, Q> {
...
+    pub fn get_mut(&mut self, entity: Entity) -> Option<Q::Item<'_>> {
+        self.view.get_mut(entity)
+    }
+
+    /// Equivalent to `get(entity).is_some()`, but does not require `Q: QueryShared`
+    ///
+    /// See [View::contains].
```
Ralith, line 1686:
```suggestion
    /// See [`View::contains`].
```

src/query.rs
```diff
@@ -1656,6 +1641,118 @@ impl<'a, 'q, Q: Query> IntoIterator for &'a mut PreparedView<'q, Q> {
...
+    pub fn contains(&self, entity: Entity) -> bool {
+        self.view.contains(entity)
+    }
+
+    /// Like `get_mut`, but allows simultaneous access to multiple entities
+    ///
+    /// See [View::get_unchecked].
```
Ralith, line 1693:
```suggestion
    /// See [`View::get_unchecked`].
```

src/query.rs
```diff
@@ -1656,6 +1641,118 @@ impl<'a, 'q, Q: Query> IntoIterator for &'a mut PreparedView<'q, Q> {
...
+    /// Must not be invoked while any unique borrow of the fetched components of `entity` is live.
+    pub unsafe fn get_unchecked(&self, entity: Entity) -> Option<Q::Item<'_>> {
+        self.view.get_unchecked(entity)
+    }
+
+    /// Like `get_mut`, but allows checked simultaneous access to multiple entities
+    ///
+    /// For N > 3, the check for distinct entities will clone the array and take O(N log N) time.
+    ///
+    /// See [View::get_many_mut].
```
Ralith, line 1706:
```suggestion
    /// See [`View::get_many_mut`].
```

src/query.rs
```diff
@@ -1656,6 +1641,118 @@ impl<'a, 'q, Q: Query> IntoIterator for &'a mut PreparedView<'q, Q> {
...
+    pub fn get_many_mut<const N: usize>(
+        &mut self,
+        entities: [Entity; N],
+    ) -> [Option<Q::Item<'_>>; N] {
+        self.view.get_many_mut(entities)
+    }
+
+    /// Iterate over all entities satisfying `Q`
+    ///
+    /// See [View::iter_mut]
```
Ralith, line 1734:
```suggestion
    /// See [`View::iter_mut`]
```

src/query.rs
```diff
@@ -1656,6 +1641,118 @@ impl<'a, 'q, Q: Query> IntoIterator for &'a mut PreparedView<'q, Q> {
     }
 }
 
+/// A borrow of a [`World`](crate::World) sufficient to random-access the results of the query `Q`.
+///
+/// Note that borrows are not released until this object is dropped.
+///
+/// This struct is a thin wrapper around [`View`](crate::View). See it for more documentation.
```
Ralith, line 1648:
```suggestion
/// This struct is a thin wrapper around [`View`]. See it for more documentation.
```
(no-op, but we might as well be concise since we rely on `View` being in scope for all the docs below)

Ralith (changes requested):

Ralith (approved):
Thanks!

sanbox-irl:
Thanks Ralith!

Issue #333, opened 2023-08-05
Is it possible to clone World? With use case.


johan-overbye:
I'm making a game editor and game with hecs. All of the game state is currently stored in a hecs `World`. The idea is the game designer sets up the initial `World` state by adding entities and components, then plays the game (prototype) in the same editor app. At that point I need to clone the `World` somehow. 

Its components contain `Entity` values (dialogue graph etc), so I can't just spawn and clone individual components. As I'm writing this I'm seeing that `insert` allows you to assign components to a specific `Entity`, so I guess I could use that? Seems inefficient though.

adamreichold:
You can (ab)use the serialization mechanism for clone by serializing and then deserializing. Admittedly, it has some drawbacks w.r.t. efficiency and determinism, but it should work right now.

Alternatively, we could implement something like the `World::clone(&self, &Cloner) -> Word` interface described in https://github.com/Ralith/hecs/issues/332#issuecomment-1651560564 where one would need to register `Clone` and `Copy` impls for all component types (until specialization stabilizes), but we get exact copies of `World` including allocator state and with minimal overhead.

johan-overbye:
Thanks @adamreichold. :) That's exactly what I ended up doing. I actually realised I'd probably want to serialise/deserialise regardless, as some of the components have "editable" versions, and I'll want the non-editable ones for play testing. For example, editable versions of components have `String`s for numeric fields so I can edit them with my immediate-mode UI, but they serialise to numbers.

In other words, would be very cool if `World::clone` were a thing, but I wouldn't actually use it for my current use case after all.

Off-topic: what would be a good serde format for this sort of thing?

adamreichold:
> Off-topic: what would be a good serde format for this sort of thing?

My personal first choice would be `bincode` as a simple and efficient codec. However, that crate is currently shifting its focus away from `serde` support. `postcard` or `dlhn` might be considered reasonable successors with focussed on `serde` support.

Ralith:
> Alternatively, we could implement something like the World::clone(&self, &Cloner) -> Word interface described in https://github.com/Ralith/hecs/issues/332#issuecomment-1651560564

It's worth noting that the public API is already powerful enough for this to be implemented downstream, even column-oriented.

> is currently shifting its focus away from `serde` support

It's also possible to use non-serde serialization, though it may take a bit of work. The out-of-the-box serde serialization code should serve as a good example; it uses only public hecs APIs.

adamreichold:
[...]

Issue #332, opened 2023-07-26
Determinism through serialization cycle


Uriopass:
I'm trying to make a replay system based on determinism. Actions on the world are saved in a replay and are applied on a `hecs::World`. `Entity` objects are stored separately in resources and serialized too.

I have a determinism problem demonstrated via the following code.
If a world has gone through a serialization cycle, the entity allocation state is not serialized and therefore entity do not necessarily get the same IDs.

```rust
let mut w1 = World::new();
let e = w1.spawn((Comp,));
w1.despawn(e);

let w2 = deserialize(serialize(&w1)); // simplified for example

let e1 = w1.spawn((Comp,));
let e2 = w2.spawn((Comp,));

assert_eq!(e1, e2); // fail! generations don't match. 
// A longer example can also show a set of operations where the id (not the genration) doesn't match either
// since the freelist is not saved.
```

From what I understand, this is intended.
What should I do if I want reproducible worlds that survive through serialization cycles?
That is, doing `operations1 -> ser -> deser -> operations2` results in the same World as `operations1 -> operations2`.

adamreichold:
What about attaching a persistent ID to each entity which could then be used to match up the new transient entity ID after deserialization? Actions would of course also have to record persistent ID of the entities they were applied to.

Uriopass:
I would find it disappointing that I must have extra state when the information is already there (entities already have unique ids). But you're right it might be more robust. I'm not sure how much magic would be required to make this work.

Uriopass:
Iteration order might also change, but I don't know enough about hecs internals to know about this.

Uriopass:
I'm using the serialization as a way to build a replay system and I want to ensure determinism, so sadly cloning wouldn't help here.

Ralith:
I recognize the usefulness of being able to save/restore allocator state for some applications. I don't think it costs us much to expose the freelist for folks to serialize if they really want; even if we have to break it in the future, the cost of doing so is small.

Would you also need generations to be assigned consistently before/after serialization? That might be a good chunk of additional data, and a bit harder to expose ergonomically.

Uriopass:
How big would it really be? It's all about saving the IDs which aren't that big, no? I'd say generation is not that expensive and ensures 100% state restoration.

Ralith:
Generations would double the amount of data, and aren't conveniently laid out in a dense slice. Not a deal breaker, but would require a bit of thought to expose gracefully.

Uriopass:
[...]

PR #372, opened 2024-05-16
Add World::try_clone()

caspark:
I would like to clone `World` (see also #333), and as far as I can tell this is currently only possible in a roundabout way by using serialization, deserialization and the determinism patch from #364 (because I want the resulting world state to be identical so that subsequent simulation is deterministic, for rollback networking purposes).

So, this PR implements making `World` clonable.

In short:

1. Users must register the types of all components used in the world into a new `Cloner` struct.
2. The `Cloner` is passed to `World::try_clone()`. Most of `World` is copied using derived `Clone` implementations, but archetypes use the `Cloner` to bulk copy component data:
    a. For `Copy` components, a bitwise copy is made of all an archetype's data.
    b. For `Clone` components, each entity is copied one by one.
    c. (There's a straightforward future extension path here of allowing custom cloning strategies to allow users to e.g. turn structural sharing on/off, but I don't need that yet so I haven't added it.)
3. If a type is not registered with the `Cloner`, an error capturing the unrecognized type is returned from `World::try_clone()`. Presumably most users will choose to unwrap it, but I figure it's nice to give folks the option of handling it.

(The implementation is based on the approach used by @adamreichold in [rs-ecs](https://github.com/adamreichold/rs-ecs), as he suggested [in a comment on #332](https://github.com/Ralith/hecs/issues/332#issuecomment-1651560564).)

I did see `DynamicClone` in `bundle.rs`, however that only clones a single instance of a type - this approach should be faster for Copy types - though I haven't done any speed tests.

I have written documentation and a few tests (& run them with miri), but I am new to this codebase and not experienced with unsafe Rust - so please check whether I have violated any invariants! Happy to adjust naming/etc as desired too.

### DIFF
--- a/src/entities.rs
+++ b/src/entities.rs
@@ -197,7 +197,7 @@ impl Clone for Entities {
             meta: self.meta.clone(),
             pending: self.pending.clone(),
             free_cursor: AtomicIsize::new(self.free_cursor.load(Ordering::Relaxed)),
-            len: self.len.clone(),
+            len: self.len,
         }
     }
 }
--- a/src/entities.rs
+++ b/src/entities.rs
@@ -191,6 +191,17 @@ pub(crate) struct Entities {
     len: u32,
 }
 
+impl Clone for Entities {
+    fn clone(&self) -> Self {
+        Self {
+            meta: self.meta.clone(),
+            pending: self.pending.clone(),
+            free_cursor: AtomicIsize::new(self.free_cursor.load(Ordering::Relaxed)),
+            len: self.len.clone(),
+        }
+    }
+}
+
 impl Entities {
     /// Reserve entity IDs concurrently
     ///
--- a/src/world.rs
+++ b/src/world.rs
@@ -949,9 +949,12 @@ impl World {
     /// assert_eq!(new1, new2);
     /// ```
     pub fn try_clone(&mut self, cloner: &Cloner) -> Result<Self, TypeUnknownToCloner> {
+        // clone archetypes first to avoid unnecessary allocations if cloning archetypes fails
+        let archetypes = self.archetypes.try_clone(cloner)?;
+
         let cloned = Self {
             entities: self.entities.clone(),
-            archetypes: self.archetypes.try_clone(cloner)?,
+            archetypes,
             bundle_to_archetype: self.bundle_to_archetype.clone(),
             insert_edges: self.insert_edges.clone(),
             remove_edges: self.remove_edges.clone(),
--- a/src/world.rs
+++ b/src/world.rs
@@ -5,7 +5,6 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-use crate::alloc::{vec, vec::Vec};
 use core::any::TypeId;
 use core::borrow::Borrow;
 use core::convert::TryFrom;
@@ -20,7 +19,9 @@ use std::error::Error;
 use hashbrown::hash_map::{Entry, HashMap};
 
 use crate::alloc::boxed::Box;
+use crate::alloc::{vec, vec::Vec};
 use crate::archetype::{Archetype, TypeIdMap, TypeInfo};
+use crate::cloning::{Cloner, TypeUnknownToCloner};
 use crate::entities::{Entities, EntityMeta, Location, ReserveEntitiesIterator};
 use crate::query::{assert_borrow, assert_distinct};
 use crate::{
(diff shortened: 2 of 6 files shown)


src/entities.rs
```diff
@@ -191,6 +191,17 @@ pub(crate) struct Entities {
     len: u32,
 }
 
+impl Clone for Entities {
+    fn clone(&self) -> Self {
+        Self {
+            meta: self.meta.clone(),
+            pending: self.pending.clone(),
+            free_cursor: AtomicIsize::new(self.free_cursor.load(Ordering::Relaxed)),
```
adamreichold, line 199: I wonder if this should be played safe and the concurrently modifiable state should not be cloned? For example, would pending updated applied to both worlds?

src/world.rs
```diff
@@ -901,6 +902,65 @@ impl World {
...
+    ///
+    /// // subsequently created entities will be allocated the same entity ids
+    /// let new1 = world1.spawn((789, "ghi".to_owned()));
+    /// let new2 = world2.spawn((012, "jkl".to_owned()));
+    /// assert_eq!(new1, new2);
+    /// ```
+    pub fn try_clone(&mut self, cloner: &Cloner) -> Result<Self, TypeUnknownToCloner> {
+        let cloned = Self {
+            entities: self.entities.clone(),
+            archetypes: self.archetypes.try_clone(cloner)?,
```
adamreichold, line 954: Maybe would be best to call this first to avoid the other work if this fails?

src/world.rs
```diff
@@ -1294,9 +1354,27 @@ impl ArchetypeSet {
...
+        Ok(Self {
+            index: self.index.clone(),
+            archetypes: {
+                let mut vec = Vec::with_capacity(self.archetypes.capacity());
+                for archetype in self.archetypes.iter_mut() {
+                    let cloned = archetype.try_clone(cloner)?;
+                    vec.push(cloned)
+                }
+                vec
+            },
```
adamreichold, line 1371: Something like

```suggestion
            archetypes: self.archetypes
                .iter()
                .map(|archetype| archetype.try_clone(cloner))
                .collect::<Result<Vec<_>, _>>()?,
```

should also work AFAIU.

Ralith:
Have you tried implementing this in terms of the public API instead, like the example serialization implementations do? The same primitives used to support column-major serialization should be suitable for efficient bulk copies.

src/world.rs
caspark: Ah, I need iter_mut(), but yes, you're right - no need to manually set capacity & loop here. Fixed.

src/world.rs
caspark: It seems like a pretty niche case for a user to care about, but: sure, done.

src/entities.rs
adamreichold: I agree on the `free_cursor` and I do not fear undefined behaviour here, but rather surprising behaviour if there are pending updates which still need to be applied.

> Perhaps it might be worth making World::try_clone() error if the World needs to be flushed (probably via replacing Entities::clone() with Entities::try_clone(), on the assumption that cloning a world that needs to be flushed is probably a mistake on the part of the user? (I don't understand exactly when World::flush() (which is pub) would be expected to be called by the user, so I can't really judge whether there's a valid use case for cloning an unflushed world)

Yes, I think erring out if the world has pending changes (and documenting that requirement) would be a more ergonomic API.

caspark:
Pushed some more changes that should make CI go green. (Embarrassingly, I had run miri and format - but then made more changes and forgot to run them again!)

Re using the column based APIs: initially when I tried that path I didn't understand enough of how hecs worked yet, but I just tried again. Here's my implementation of cloning World using the public column-based APIs, suitable for copy-pasting into `examples/`: https://gist.github.com/caspark/62a092d3068ffba7781008b2828b5320

Shortcomings of using the column-based APIs (ordered from worst to okay):

1. Entity allocator state is not preserved, so entity generations and entity allocation order can be different in the clone. Could potentially be worked around by implementing this in `impl World`,  removing the logic to reserve entities in the cloned world and duplicating-much-of `World::spawn_column_batch()` into the clone function for `World`? I am not sure if that would work because I think (in theory) `ArchetypeSet`'s hashmap might end up with different ordering.
2. To copy a slice of `Copy` types in one go (instead of instance-by-instance), we'd need to add something like `BatchWriter::extend(&mut, slice: &[T]) where T: Copy`, which uses `ptr::copy_nonoverlapping()`. That looks doable, but `BatchWriter` currently operates in terms of `IterMut<MaybeUninit<T>>`  - so I think that'd have to be rewritten to deal with pointers directly? (which'd still be some new/changed unsafe code, though likely a smaller chunk than `Archetype::try_clone()` in this PR).
  3. Perhaps the current implementation in the gist does actually compile down to copying from the slice directly? Hard to say.
4. Since that approach operates on static types (rather than "dynamic" types as in this PR), each type needs to be specified directly, and specified again a second time. That's not a big deal in and of itself (can always provide/write a macro) but it slightly complicates fixing `1.` above: if preserving entity allocator state means moving the gist's `clone_world` into `impl World` then that function would need to accept a `CloningContext` like the column based serialization does.

Overall, it seems like it would work with sufficient poking and prodding, but I am not sure whether it's a clear win or not: it'd have some unsafe code too, and preserving entity allocator state in that type of approach seems harder (both to implement for me, and for you as a guarantee to maintain into the future).

What do you think @Ralith ?

Ralith:
> Entity allocator state is not preserved, so entity generations and entity allocation order can be different in the clone.

This is an unrelated problem, and one we should address in separate work, e.g. https://github.com/Ralith/hecs/pull/364.

> To copy a slice of Copy types in one go (instead of instance-by-instance), we'd need to [...]

LLVM is very good at optimizing memcpy-like loops into a literal memcpy or similarly efficient code. Handling `Copy` types separately in the absence of disassembly and benchmarking is a premature optimization, especially if it complicates the API

> Since that approach operates on static types (rather than "dynamic" types as in this PR), each type needs to be specified directly, and specified again a second time.

This can be easily avoided using type erasure, much like you do in the existing PR. Something like:

```rs
struct WorldCloner {
    registry: TypeIdMap<&'static dyn Fn(&Archetype, &mut ColumnBatchBuilder)>,
}

impl WorldCloner {
    pub fn register<T: Component + Clone>(&mut self) {
        self.registry.insert(TypeId::of::<T>(), &|src, dest| {
            let mut column = dest.writer::<T>().unwrap();
            for component in &*src.get::<&T>().unwrap() {
                _ = column.push(component.clone());
            }
        });
    }
}
```

caspark:
Sorry for falling off the face of the earth for a while - I've been busy using my fork for my project so responding here kept sliding down the todo list.

1. Re tackling preserving entity allocator state in a separate PR - sure, that seems reasonable.
2. Re handling copy types separately being a premature optimization - okay, yep, fair.
3. This seems plausible, though some private types need to be exposed and `ColumnBatchType` needs to gain `add_dynamic()`.

I'll close this PR and raise a new PR to implement cloning as an example using the public API.

PR #375, opened 2024-09-12
Allow multiple trackers for a single component type

Azkellas:
This PR introduces a flag type in `ChangeTracker` (from `ChangeTracker<Component>` to `ChangeTracker<Component, Flag>`) to allow multiple trackers for a single component type.

I wanted to track a component type from different areas of my project, so I added this change. This is obviously a breaking change.

On a related note, it would be interesting to track removed components from deleted entites, though that's a difficult task to tackle.

Thanks for this great crate!

### DIFF
--- a/src/change_tracker.rs
+++ b/src/change_tracker.rs
@@ -1,6 +1,7 @@
 use core::mem;
 
 use alloc::vec::Vec;
+use core::marker::PhantomData;
 
 use crate::{Component, Entity, PreparedQuery, With, Without, World};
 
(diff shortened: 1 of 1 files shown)


Ralith:
Thanks for the PR! I recommend either using a single tracker and broadcasting its information to all interested parties, or vendoring the change tracker implementation into your project and making the changes yourself. Note that the implementation uses only public hecs APIs, so this should be straightforward.

There's a lot of different ways people might want to track changes or route that information, and I don't think it's feasible to support them all out of the box, so this implementation is really just intended to get you started and/or for simple cases.

Issue #380, opened 2024-10-17
Example for serializing, deserializing and using the `SerializeContext`?


kaphula:
Hey, 

Do we have any working examples for demonstrating de/serialization in the context of saving world to disk and loading it back?

Ralith:
There are no examples beyond the rustdoc, which should cover the specifically tricky parts. I'm open to merging a clean, commented end-to-end example, if you like.

Ralith:
Sounds like you understand it fine to me. Did you have trouble doing that?

kaphula:
Alright, thanks! I did not try it yet because I did not want to waste time implementing something I did not understand.

PR #381, opened 2024-10-17
Add World serialization example

kaphula:
As discussed in https://github.com/Ralith/hecs/issues/380

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 **/*.rs.bk
 Cargo.lock
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 /target
 **/*.rs.bk
 Cargo.lock
-.idea
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,3 +53,8 @@ debug = true
 
 [workspace]
 members = ["macros", "tests/no-std-test-crates/macros"]
+
+
+[[example]]
+name = "serialize_to_disk"
+required-features = ["column-serialize"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ rand = "0.8.3"
 trybuild = "1.0.23"
 serde = { version = "1.0.117", features = ["derive"] }
 serde_test = "1.0.117"
+serde_json = "1.0.117"
 
 [[bench]]
 name = "bench"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ rand = "0.8.3"
 trybuild = "1.0.23"
 serde = { version = "1.0.117", features = ["derive"] }
 serde_test = "1.0.117"
-serde_json = "1.0.117"
+bincode = "1.3.3"
 
 [[bench]]
 name = "bench"
--- a/examples/serialize_to_disk.rs
+++ b/examples/serialize_to_disk.rs
@@ -1,0 +1,177 @@
+//! This example demonstrates how to save [hecs::World] instance
+//! to disk and load it back to memory using serialization. It can be useful for implementing
+//! a save mechanism for a game.
+//!
+//! The example creates a sample `World`, serializes it, saves it to disk as `saved_world.world` file,
+//! loads it back from the disk, deserializes the loaded world data and validates that component
+//! data of the worlds match.
+//!
+//! Run this example from crate root with:
+//! `cargo run --example serialize_to_disk --features "column-serialize"`
+
+#[cfg(feature = "column-serialize")]
+mod serialize_to_disk_example {
+   use std::any::TypeId;
+   pub use std::fs::File;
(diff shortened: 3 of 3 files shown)


examples/serialize_to_disk.rs
```diff
@@ -0,0 +1,186 @@
...
+//! a save mechanism for a game.
+//!
+//! The example creates a sample `World`, serializes it, saves it to disk as `saved_world.world` file,
+//! loads it back from the disk, deserializes the loaded world data and validates that component
+//! data of the worlds match.
+//!
+//! Run this example from crate root with:
+//! `cargo run --example serialize_to_disk --features "column-serialize"`
+
+#[cfg(feature = "column-serialize")]
```
Ralith, line 12: I'm not sure what the point of this cfg gate is, because this will fail to compile regardless if the feature is disabled. Instead, define the feature as required for the example in Cargo.toml.

examples/serialize_to_disk.rs
```diff
@@ -0,0 +1,186 @@
...
+mod serialize_to_disk_example {
```
Ralith, line 13: Why have an inner module here?

Cargo.toml
```diff
@@ -41,6 +41,7 @@ rand = "0.8.3"
 trybuild = "1.0.23"
 serde = { version = "1.0.117", features = ["derive"] }
 serde_test = "1.0.117"
+serde_json = "1.0.117"
```
Ralith, line 44: JSON is a poor choice to illustrate column-major serialization. If you want the output to be human-readable, use row-major serialization. If you don't, don't use a slow, complex, text-based format.

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 **/*.rs.bk
 Cargo.lock
+.idea
```
Ralith, line 4: Unrelated change.

Ralith (changes requested):

Cargo.toml
kaphula: Okay, should I use bincode instead or something else?

examples/serialize_to_disk.rs
kaphula: True!

Cargo.toml
Ralith: If you want to illustrate columnar serialization, bincode would be a reasonable choice.

kaphula:
I applied the fixes you suggested. Does it look better now?

examples/serialize_to_disk.rs
Ralith: This comment has not been addressed.

examples/serialize_to_disk.rs
```diff
@@ -0,0 +1,172 @@
...
+                ComponentId::ComponentB => {
+                    deserialize_column::<ComponentB, _>(entity_count, &mut seq, batch)?;
+                }
+            }
+        }
+        Ok(())
+    }
+}
+
+impl SerializeContext for SaveContext {
```
Ralith, line 93: This shouldn't be implemented on the same type as `DeserializeContext`, because it doesn't need the same (or any) data on the side.

Ralith (changes requested):

examples/serialize_to_disk.rs
kaphula: So should it be done like this?

```rust
#[derive(Default)]
pub struct SaveContextSerialize {}
#[derive(Default)]
pub struct SaveContextDeserialize {
   pub components: Vec<ComponentId>,
}
```

Ralith: That would be clearer, yes.

examples/serialize_to_disk.rs
```diff
@@ -45,7 +47,7 @@ pub struct ComponentB {
 }
 
 #[cfg(feature = "column-serialize")]
```
Ralith, line 49: Still got a stray cfg.

examples/serialize_to_disk.rs
kaphula: Is it fine now?

examples/serialize_to_disk.rs
kaphula: Removed.

examples/serialize_to_disk.rs
```diff
@@ -0,0 +1,180 @@
...
+            // the same:
+            println!("Loaded world \'{}\' from disk.", path.display());
+
+            print!("Validating world data... ");
+            for (_, t) in &mut world.query::<&ComponentA>() {
+                assert_eq!(t, &input_data1);
+            }
+            for (_, t) in &mut world.query::<&ComponentB>() {
+                assert_eq!(t, &input_data2);
+            }
```
Ralith, line 173: You should also check `world.len()`, or this will pass if the world is empty.

Ralith (approved):

Ralith:
Thanks!

Issue #376, opened 2024-09-18
OneOf Query Type


Moulberry:
Hello, hope you are doing well.

This issue is about the discussion of a potential new query type OneOf.
I am willing to create the pull request for this feature if it's desired, I'm creating an issue first to gauge whether or not this would be suitable for hecs.

## Motivation

I can't find a built-in way to query one of a set of types, the closest is the Or query type but this has certain limitations as discussed below. The motivation for OneOf would be to implement different behaviour when the player interacts with some entity, i.e. clicking an Enemy/Item/Corpse/etc. 

## Example
The query type `OneOf3<A, B, C>` will match any entity that contains either A, B or C, but only one of them.

You can then use code similar on the result:
```rust
match item {
    One(a) => {},
    Two(b) => {},
    Three(c) => {},
}
```

## Alternatives that already exist (to my knowledge)

Layering `Or`s, like `Or<A, Or<B, C>>`
Problems:
1. Lots of boilerplate to set up the query with the repeated or and brackets
2. Not very easy to concisely match on the result
3. The query engine will continue checking for B and C even if A is present, i.e. it doesn't stop when you've already matched

Using a lot of options, like `(Option<A>, Option<B>, Option<C>)`
Problems:
1. Code using the query result will end up having a lot of if-else chains to check every option
2. 3 `None`s is a valid result, meaning you're going to end up iterating over every entity unless you constrain it with a more complex query

Some other workaround, like inserting an Attack component when an entity is clicked
Problems:
1. Doesn't work for every use-case
2. Performance penalty due to changing archetype
3. More architecturally complex than just performing the logic you want directly

I apologize in advance if there's built-in a way to do OneOf. Once again, I am willing to contribute this feature if it's within the scope of hecs. Thank you for your time

Moulberry:
> Permanently add
> 
> ```
> enum ClickTarget {
> Enemy,
> Item,
> Button,
> }
> ```
> 
> Component and query for (Entity, ClickTarget), then match click_target and get component you want using entity_id?

A few problems:
1. This solution requires multiple queries
2. No architectural guarantees that the ClickTarget is correct. The ECS allows you to add a ClickTarget to an entity that doesn't match and it also allows you to forget to add ClickTarget to an entity that would match
3. Whether or not you want to target something is often more complex than what is expressible with a single enum variant. With this solution you'd need to either update/remote the ClickTarget or do additional checks to make sure it actually matches the query you want.
[...]

PR #384, opened 2024-11-18
Implement #[derive(Query)] for enum types

Moulberry:
#[derive(Query)] can now be applied to enum types

An Enum query will yield the first variant that satisfies an Entity (even if there are other variants which might have satisfied it)
Supports tuple enums, struct enums and will ignore any unit enum field.

Closes https://github.com/Ralith/hecs/issues/376


macros/src/query.rs
```diff
@@ -193,6 +203,327 @@ pub fn derive(input: DeriveInput) -> Result<TokenStream2> {
...
+            syn::Fields::Named(_) => quote! {
+                #ident {
+                    #(
+                        #fields: #fetches,
+                    )*
+                },
+            },
+            syn::Fields::Unnamed(_) => quote! {
+                #ident(#(#fetches),*),
+            },
```
Ralith, line 285: Could we combine these two branches in every case by using the `{0: x, 1: y, ...}` constructor/pattern syntax for tuple variants? That might save a *lot* of duplication.

Ralith (changes requested):
Thanks, this is great work!

Ralith:
Don't forget to update the rustdoc for `derive_query` in macros/src/lib.rs, also.

macros/src/query.rs
```diff
@@ -193,6 +203,327 @@ pub fn derive(input: DeriveInput) -> Result<TokenStream2> {
...
+            // Ignore empty variants since otherwise they would always match
+            continue;
+        }
+
+        let fetches = queries
+            .iter()
+            .map(|ty| quote! { <#ty as ::hecs::Query>::Fetch })
+            .collect::<Vec<_>>();
+
+        if dangling_variant.is_none() {
```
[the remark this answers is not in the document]
Moulberry: Will do, however the change won't make empty enums work since the Derive macro checks for a lifetime and (afaik) having a lifetime on an empty enum is a compile error from Rust itself

macros/src/query.rs
```diff
@@ -193,6 +203,327 @@ pub fn derive(input: DeriveInput) -> Result<TokenStream2> {
...
+            },
+            syn::Fields::Unit => unreachable!(),
+        });
+
+        let intermediates = fields
+            .iter()
+            .map(|x| match x {
+                syn::Member::Named(ref ident) => ident.clone(),
+                syn::Member::Unnamed(ref index) => {
+                    Ident::new(&format!("field_{}", index.index), Span::call_site())
```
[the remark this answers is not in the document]
Moulberry: Should I update all other usages of call_site() including those in the original code for deriving structs?

macros/src/query.rs
```diff
@@ -193,6 +203,327 @@ pub fn derive(input: DeriveInput) -> Result<TokenStream2> {
...
+                    )
+                })
+                .unzip(),
+            syn::Fields::Unit => continue,
+        };
+
+        let ident = variant.ident.clone();
+
+        if fields.is_empty() {
+            // Ignore empty variants since otherwise they would always match
```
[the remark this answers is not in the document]
Moulberry: [fc0236c](https://github.com/Ralith/hecs/pull/384/commits/fc0236cfafe5e0427ab39c4cc543156d89bfc6e2)

macros/src/query.rs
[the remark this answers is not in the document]
Moulberry: [fc0236c](https://github.com/Ralith/hecs/pull/384/commits/fc0236cfafe5e0427ab39c4cc543156d89bfc6e2)

macros/src/query.rs
```diff
@@ -193,6 +203,327 @@ pub fn derive(input: DeriveInput) -> Result<TokenStream2> {
...
+                    )*
+                    Self::Item::#ident { #(#intermediates,)* }
+                },
+            },
+            syn::Fields::Unnamed(_) => quote! {
+                Self::Fetch::#ident(#(#intermediates),*) => {
+                    #(
+                        let #intermediates: <#queries as ::hecs::Query>::Item<'q> = <#queries as ::hecs::Query>::get(#intermediates, n);
+                    )*
+                    Self::Item::#ident(#(#intermediates,)*)
```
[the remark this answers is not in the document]
Moulberry: A similar approach way used by the derive macro for the struct. It seems like their purpose is to assert that the lifetimes are valid. Without the intermediate variables, you can create a query with a static lifetime resulting in undefined behaviour. The wrong_lifetime.rs test exists to catch this mistake

macros/src/query.rs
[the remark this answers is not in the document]
Moulberry: [1b61148](https://github.com/Ralith/hecs/pull/384/commits/1b61148446c447a55bf003ea19e74b9b17be9d57)

macros/src/query.rs
Moulberry: It wasn't possible to combine every branch in that way.
Instead, I converted the State and Fetch enums to always be struct enums which eliminates almost all of the duplication.
[39209c6](https://github.com/Ralith/hecs/pull/384/commits/39209c68e04e53a37fb0e3ffbd35ecb985d71fa2)

macros/src/query.rs
```diff
@@ -193,6 +203,327 @@ pub fn derive(input: DeriveInput) -> Result<TokenStream2> {
...
+            })
+            .collect::<Vec<_>>();
+
+        query_get_variants.extend(match variant.fields {
+            syn::Fields::Named(_) => quote! {
+                Self::Fetch::#ident { #(#intermediates),* } => {
+                    #(
+                        let #intermediates: <#queries as ::hecs::Query>::Item<'q> = <#queries as ::hecs::Query>::get(#intermediates, n);
+                    )*
+                    Self::Item::#ident { #(#intermediates,)* }
```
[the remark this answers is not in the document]
Moulberry: Done in [39209c6](https://github.com/Ralith/hecs/pull/384/commits/39209c68e04e53a37fb0e3ffbd35ecb985d71fa2)

macros/src/lib.rs
```diff
@@ -63,12 +63,15 @@ pub fn derive_dynamic_bundle_clone(input: TokenStream) -> TokenStream {
...
-/// Implement `Query` for a struct
+/// Implement `Query` for a struct or enum.
 ///
-/// Queries structs can be passed to the type parameter of `World::query`. They must have exactly
+/// Queries can be passed to the type parameter of `World::query`. They must have exactly
 /// one lifetime parameter, and all of their fields must be queries (e.g. references) using that
 /// lifetime.
 ///
+/// For enum queries, the result will always be the first variant that satisfies the entity.
+/// Unit variants and variants without any fields will always satisfy an entity.
```
Ralith, line 73: An entity satisfies a query, not the reverse.

macros/src/query.rs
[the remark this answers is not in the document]
Ralith: Ah, it's not the intermediate variables per se that matter, but the type annotation 👍

macros/src/query.rs
```diff
@@ -193,14 +208,276 @@ pub fn derive(input: DeriveInput) -> Result<TokenStream2> {
...
+                Self::Fetch::#ident { #(#named_fields),* } => {
+                    #(
+                        let #named_fields: <#queries as ::hecs::Query>::Item<'q> = <#queries as ::hecs::Query>::get(#named_fields, n);
+                    )*
+                    Self::Item::#ident { #( #fields: #named_fields,)* }
+                },
+            },
+            syn::Fields::Unit => quote! {
+                Self::Fetch::#ident {} => Self::Item::#ident,
+            },
```
Ralith, line 314: Do these need to be separate branches at all? If `named_fields` is empty, the first branch would generate `Self::Fetch::#ident {} => Self::Item::#ident {}`, which should be perfectly legal.

tests/derive/enum_query.rs
```diff
@@ -0,0 +1,24 @@
...
+#[derive(Query)]
+enum All<'a> {
+    Foo(&'a i32),
+    Bar {
+        bar: &'a bool
+    },
+    Baz
+}
+
+fn main() {}
```
Ralith, line 24: nit: missing trailing newline

tests/derive/enum_unsupported.stderr
```diff
@@ -0,0 +1,5 @@
+error: derive(Bundle) does not support enums or unions
```
Ralith, line 1: Misleading error message

Ralith (changes requested):

tests/derive/enum_unsupported.stderr
Moulberry: That error message is for #[derive(Bundle)], which still only support structs
This PR only adds support for #[derive(Query)]

macros/src/query.rs
Moulberry: When matching, struct enum variants would use `Enum::StructVariant { a, b, c }` whereas tuple enum variants would use `Enum::EnumVariant {0: a, 1: b, 2: c}`. It probably would have been possible to generate a TokenStream called `match_args` but converting everything into a struct enum was less lines of code anyways.

macros/src/lib.rs
Moulberry: Does "match" work instead? [c31467e](https://github.com/Ralith/hecs/pull/384/commits/c31467e7b93fdfc07818812ada266eb510b2e48d)

macros/src/query.rs
Moulberry: Ah, I didn't know you could construct a unit variant with empty braces.
[27b9eea](https://github.com/Ralith/hecs/pull/384/commits/27b9eea26af3077d6139c04fe60562d8fe1c8902)

tests/derive/enum_query.rs
Moulberry: [7d48c02](https://github.com/Ralith/hecs/pull/384/commits/7d48c025235f1820fe68c973e6f6f733ba83b803)

tests/derive/enum_unsupported.stderr
Ralith: Whoops, right.

macros/src/lib.rs
Ralith: Sure.

Ralith (approved):

PR #390, opened 2025-03-09
Expose freelist accessors for deterministic Entity allocation

Ralith:


caspark:
Thanks for working on this! Haven't tried this branch out in my game yet but it seems to work fine in the updated example over at #394.

src/entities.rs
```diff
@@ -529,6 +532,45 @@ impl Entities {
...
+                };
+                assert_eq!(
+                    meta.location.index,
+                    u32::MAX,
+                    "freelist addresses live entities"
+                );
+            }
+        }
+        if let Some(max) = freelist.iter().map(|e: &Entity| e.id()).max() {
+            if max as usize > self.meta.len() {
```
caspark, line 563: Should be `>=` to avoid out of bounds array access at `self.meta[entity.id as usize].generation = entity.generation;` below.

Minimal test case you might or might not want to add:

```rust

    #[test]
    fn test_set_freelist_causes_resize() {
        let mut world = World::new();
        let a = world.spawn(());
        world.despawn(a).unwrap();
        let freelist = world.freelist().collect::<Vec<_>>();

        let mut world2 = World::new();
        world2.set_freelist(&freelist);
    }
```

src/world.rs
```diff
@@ -912,6 +912,31 @@ impl World {
...
+    /// freelist.
+    pub fn freelist(&self) -> impl ExactSizeIterator<Item = Entity> + '_ {
+        self.entities.freelist()
+    }
+
+    /// Designate entity IDs to be reused by future spawns
+    ///
+    /// See [`freelist`](Self::freelist). May only be called when the set of
+    /// live entity IDs exactly matches those that were live when the supplied
+    /// freelist was captured.
```
caspark, line 936: I was thinking about how Entities clears its `pending` vec without calling `flush` and whether that's ok in weird cases like `deserialize, spawn, despawn, reserve entity, set freelist, use reserved entity ID` - and I think it is because this second sentence here is supposed to rule out any case where the `pending` vec matters.

But it might be worth spelling it out with another docs sentence? Something like `For example, spawning, despawning, or reserving an entity ID would invalidate a previously captured freelist.`

src/entities.rs
Ralith: Great catch, thanks!

PR #394, opened 2025-03-12
Update clone example to preserve entity allocator state

caspark:
Tested out #390 by updating the clone example - seems to work as expected in this test, though I had to do a bit of trial and error to realize that it was important to set the freelist _after_ doing the batch column spawning.

Would be good to make a similar update for (at least) column-wise serialization too, but I'm out of time for now :)

### DIFF
--- a/examples/cloning.rs
+++ b/examples/cloning.rs
@@ -1,14 +1,12 @@
 //! This example demonstrates using the [ColumnBatch][hecs::ColumnBatch] API to efficiently clone
 //! the entities in a [World] along with some or all components.
 //!
-//! Note that the cloned world may have different iteration order and/or newly created entity ids
-//! may diverge between the original and newly created worlds. If that is a dealbreaker for you,
-//! see https://github.com/Ralith/hecs/issues/332 for some pointers on preserving entity allocator
-//! state; as of time of writing, you'll need to patch `hecs`.
-
-use std::any::TypeId;
+//! By using [`World::set_freelist()`], we ensure that the cloned world will allocate new entities
+//! in sync with the original world: newly allocated entities will have the same entity ids as the
+//! original world, and the iteration order will be the same.
 
 use hecs::{Archetype, ColumnBatchBuilder, ColumnBatchType, Component, TypeIdMap, TypeInfo, World};
+use std::any::TypeId;
 
 struct ComponentCloneMetadata {
     type_info: TypeInfo,
@@ -45,6 +43,8 @@ impl WorldCloner {
     }
 
     fn clone_world(&self, world: &World) -> World {
+        let freelist = world.freelist().collect::<Vec<_>>();
+
         let mut cloned = World::new();
 
         for archetype in world.archetypes() {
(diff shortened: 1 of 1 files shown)


examples/cloning.rs
```diff
@@ -1,14 +1,12 @@
...
 //!
-//! Note that the cloned world may have different iteration order and/or newly created entity ids
-//! may diverge between the original and newly created worlds. If that is a dealbreaker for you,
-//! see https://github.com/Ralith/hecs/issues/332 for some pointers on preserving entity allocator
-//! state; as of time of writing, you'll need to patch `hecs`.
-
-use std::any::TypeId;
+//! By using [`World::set_freelist()`], we ensure that the cloned world will allocate new entities
+//! in sync with the original world: newly allocated entities will have the same entity ids as the
+//! original world, and the iteration order will be the same.
```
Ralith, line 6: I don't think this is enough to guarantee consistent iteration order. IIRC that depends on the order in which archetypes are introduced, which might still vary.

Ralith (changes requested):

examples/cloning.rs
Ralith: Actually, this probably does work because the example is saving/restoring column batches in order. Probably merits calling that out explicitly.

Ralith:
> I had to do a bit of trial and error to realize that it was important to set the freelist after doing the batch column spawning.

I agree this is a bit obscure. I'm interested in any ideas you might have for how to document this better and/or produce clearer errors. As far as I can tell there's no good way to relax the requirement, since setting it *before* spawning entities makes the allocator confused about whether entities not in the freelist are allocated or not, which leads to panics.

examples/cloning.rs
caspark: Done, by updating the docs and adding a comment on the impl explaining why iteration order is incidentally preserved by this approach.

caspark:
> I agree [the req on when to set the freelist] is a bit obscure. I'm interested in any ideas [to improve]

Alright, I just read your docs again and I can only conclude I must have been afflicted by temporary blindness: your docs are very clear that the freelist should be set _after_ spawning the entities. So I don't think there's much else that can be done there to improve that.

I've addressed your comment regarding the iteration order by documenting why iteration order happens to be preserved, but to be honest it feels wrong to keep that tucked away in a comment in an example. How do you feel about committing to keeping the iteration order consistent via this approach, and enforcing that via some tests? The specific guarantees asserted would be something like:

1. `archetypes()` returns archetypes in the order in which they were added to the world, with the empty archetype always first. (Looks like there is already a test for this (`columnar_access`), though the behavior isn't documented in `archetypes()`.)
2. `iter()`/`query()` each visit entities in order of matching archetypes first, and (I assume) by ascending entity id within each archetype. (Looks like tests are currently written specifically to avoid relying on iteration order?)

Or perhaps it would be better to move `WorldCloner` into hecs proper, so that its behavior of preserving iteration order can be directly verified in a test without having to lock down the iteration order of the underlying APIs? I lean towards this approach, but I also know you are (or were?) quite keen to keep the hecs API quite thin, i.e. free of functionality which users could implement themselves via already publicly available APIs.

Or we can leave it as-is: given how stable hecs has been in the year I've used it (thanks, by the way!), I am personally not too worried about you changing the underlying implementation.

Ralith:
> The specific guarantees asserted would be something like:

Those are much stronger than necessary. We should only guarantee that iteration order is preserved when a new world is created with column batches of the same order and contents as the original world's `archetypes()`.

caspark:
I've added a test along those lines now.

CI is failing but the failures are due to Rust's 1.86 release rather than anything in this PR.

Issue #399, opened 2025-05-17
Query caching via PreparedQuery is a big speedup for empty queries — should hecs always cache queries?


LPGhatguy:
I found that preparing queries that match no entities were eating a lot of time in our game. For reference, we run on the order of 200 queries per frame. Here is the timing of some of our fastest systems. There is a big cliff between the systems that don't query the world versus those that do:

![Image](https://github.com/user-attachments/assets/97642def-7231-4f58-88b5-426a1a3ab50d)

I tried refactoring one of these systems to use `PreparedQuery` instead and saw a speedup of around 15x when that query matches no entities. Here is the before and after code of the system: https://gist.github.com/LPGhatguy/dcec605c3b6d058332f5ae527af6efde

...so naturally I thought I'd try caching all of my queries automatically! I implemented a quick bodge around `World::query` to do that caching: https://gist.github.com/LPGhatguy/2c6180540e45c6fd411456af560f9d64

The speedup on the fastest systems was a lot and I saved around 1.5ms per frame! Here's the updated chart of the fastest systems for comparison:

![Image](https://github.com/user-attachments/assets/fb6619b7-1654-46ee-ae4a-331493769f19)

I imagine that a hecs-native solution could be faster, use less janky unsafe, and then would apply to everyone who uses the library. I've found that this change is a very large speedup and maybe it would be for other games too!

adamreichold:
> ...so naturally I thought I'd try caching all of my queries automatically! I implemented a quick bodge around World::query to do that caching: https://gist.github.com/LPGhatguy/2c6180540e45c6fd411456af560f9d64e

I think that if hecs added native query caching, it should work similar to `rusqlite`'s query caching, i.e. be fully transparent except for the opt-in of using an e.g. `query_cached` method instead of `query`. Of course, this means to amend the query trait infrastructure to produce some kind of hash key and it probably also implies interior mutability (on the queries itself, not just the components).

Ralith:
I think we can make this 100% transparent in the existing `query` API, keying off the `TypeId` of the query itself. Maybe I'm forgetting something because it seems weird in retrospect that we aren't already doing this.

LPGhatguy:
I have a strong preference for this caching to be done in the regular `query` method. It's a big speed up and the set of queries is likely static for most users. I do not want to have to write `query_cached` by default and would rather opt out of caching if there's a bad case.

Ralith:
Okay, this is a little trickier than expected, albeit still possible:
[...]

Issue #155, opened 2021-04-11
Optimize archetype filtering


Ralith:
Currently, each query checks compatibility with each known archetype. This works pretty well in practice because the number of archetypes doesn't tend to be astronomical and checking is fast, but there should be an asymptotically better solution.

adamreichold:
Will keeping an acceleration data structure like a mapping from type ID to matching archetypes around and up-to-date really make sense when the above cost can be avoided using prepared queries?

Ralith:
Definitely questionable! The value of making query set-up faster isn't *zero*, but it's certainly a very low priority at this point, and a complex solution would be difficult to justify.

adamreichold:
> a complex solution would be difficult to justify.

What I am mostly concerned about is that this might add overhead to component insertion/removal to speed up query set-up that even users of prepared queries who do not gain anything would have to pay. (The ordered-list-seems-faster-than-hash-table thing within the archetypes also seems to suggest than typical usage is dominated by the constants instead of the asymptotics.)

Ralith:
Average-case component insertion/removal would bypass any acceleration structure updates since target archetype indexes are cached in `ArchetypeSet::insert_edges` and `Archetype::remove_edges`.

kettle11:
One potential approach is to store a sparse-set of associated `Archetype`s per component. 

Then finding matching `Archetype`s for a query becomes the following steps:
* Find the component in the query with the fewest associated `Archetype`s
* Iterate matching `Archetype`s of that component
* For every other component's `Archetype`s in the query check its sparse-set to see if it contains that `Archetype`

This should be pretty quick.

Downsides:
 * There's a little bit of book-keeping when a new `Archetype` is created
 *  Each component's sparse-set could contain a flag per-`Archetype`. More memory usage. 
 * More code to maintain. I previously implemented the described approach in a different ECS in ~190 lines of code.

Ralith:
That sounds like a plausible approach. Memory/archetype setup CPU cost should be insignificant. To justify the maintenance/complexity cost, I'd want to see a compelling and realistic benchmark, ideally something from a real game.

Ralith:
This problem is closely related to the archetype graph we use to optimize component insertion/removal: the set of archetypes traversed by a query that accesses a certain set of components is equal to the set of archetypes reachable by adding components to an entity that initially has exactly those components.

[...]

PR #400, opened 2025-05-17
Cache queries automatically

Ralith:
Fixes https://github.com/Ralith/hecs/issues/399, https://github.com/Ralith/hecs/issues/155.

Remaining work:
- Use the cache to drive borrows
- Validate with benchmarks
- Expand use of the cache to other query methods
- Cache views too?

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -730,7 +730,7 @@ impl<'q, Q: Query> IntoIterator for &'q mut QueryBorrow<'_, Q> {
 /// Iterator over the set of entities with the components in `Q`
 pub struct QueryIter<'q, Q: Query> {
     world: &'q World,
-    archetypes: core::ops::Range<usize>,
+    archetypes: ArchetypeIter<Q>,
     iter: ChunkIter<Q>,
 }
 
@@ -740,24 +740,11 @@ impl<'q, Q: Query> QueryIter<'q, Q> {
     /// `'q` must be sufficient to guarantee that `Q` cannot violate borrow safety, either with
     /// dynamic borrow checks or by representing exclusive access to the `World`.
     unsafe fn new(world: &'q World) -> Self {
-        let n = world.archetypes().len();
         Self {
             world,
-            archetypes: 0..n,
+            archetypes: ArchetypeIter::new(world),
             iter: ChunkIter::empty(),
         }
-    }
-
-    /// Advance query to the next archetype
-    ///
-    /// Outlined from `Iterator::next` for improved iteration performance.
-    fn next_archetype(&mut self) -> Option<()> {
-        let archetype = self.archetypes.next()?;
-        let archetype = unsafe { self.world.archetypes_inner().get_unchecked(archetype) };
-        let state = Q::Fetch::prepare(archetype);
-        let fetch = state.map(|state| Q::Fetch::execute(archetype, state));
-        self.iter = fetch.map_or(ChunkIter::empty(), |fetch| ChunkIter::new(archetype, fetch));
-        Some(())
     }
 }
 
@@ -772,7 +759,10 @@ impl<'q, Q: Query> Iterator for QueryIter<'q, Q> {
         loop {
             match unsafe { self.iter.next() } {
                 None => {
-                    self.next_archetype()?;
+                    // Safety: `self.world` is the same one we passed to `ArchetypeIter::new` just above
+                    unsafe {
+                        self.iter = self.archetypes.next(self.world)?;
+                    }
                     continue;
                 }
                 Some((id, components)) => {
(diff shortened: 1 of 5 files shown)


src/query.rs
```diff
@@ -752,11 +751,7 @@ impl<'q, Q: Query> QueryIter<'q, Q> {
     ///
     /// Outlined from `Iterator::next` for improved iteration performance.
```
adamreichold, line 758: Does outlining still make sense here (rather than at `ArchetypeIter::next`)?

src/query.rs
```diff
@@ -1794,6 +1783,40 @@ fn release_borrow<Q: Query>(archetypes: &[Archetype]) {
...
+        }
+    }
+
+    fn entity_len(&self, world: &World) -> usize {
+        self.archetypes
+            .clone()
+            .map(|x| unsafe { world.archetypes_inner().get_unchecked(x) })
+            .filter(|&x| Q::Fetch::access(x).is_some())
+            .map(|x| x.len() as usize)
+            .sum::<usize>()
```
adamreichold, line 1903: Type ascription is probably not necessary any more due to signature.

src/query.rs
```diff
@@ -1794,6 +1783,40 @@ fn release_borrow<Q: Query>(archetypes: &[Archetype]) {
...
+
+impl<Q: Query> ArchetypeIter<Q> {
+    fn new(world: &World) -> Self {
+        Self {
+            archetypes: 0..world.archetypes().len(),
+            _marker: PhantomData,
+        }
+    }
+
+    fn next(&mut self, world: &World) -> Option<ChunkIter<Q>> {
```
adamreichold, line 1886: It isn't API, but I wonder if this needs to be `unsafe` as passing the wrong `world` could mean the `get_unchecked` is out of bounds?

src/query.rs
```diff
@@ -1783,11 +1789,92 @@ fn release_borrow<Q: Query>(archetypes: &[Archetype]) {
     }
 }
 
+#[cfg(feature = "std")]
+pub(crate) type QueryCache = RwLock<TypeIdMap<Arc<dyn Any + Send + Sync>>>;
```
adamreichold, line 1793: If these locks are short-lived and this is the only reason to require `std`, I would still suggest looking into using an "atomic cell" or some other form of borrow checking, especially if reentracy is controlled by this not being public API.

src/query.rs
Ralith: Good catch, thanks.

src/query.rs
Ralith: How could we gracefully expand a `TypeIdMap` inside an atomic cell? Inserting requires unique access.

adamreichold: Using an API like [`AtomicRefCell`](https://docs.rs/atomic_refcell/latest/atomic_refcell/struct.AtomicRefCell.html), i.e. `RefCell` implemented using atomics. I use it extensively to ensure cooperatively scheduled futures (which do not need blocking) remain `Send`.

Ralith: From the documentation of that crate:
> for consumers that can guarantee that they will never mutably borrow the contents concurrently with immutable borrows, an RwLock is overkill

We cannot make such a guarantee.

Ralith: Propagating that requirement to callers would prevent overlapping read-only queries, which is a major regression.

adamreichold: You are right, you still need exclusive access for compatible queries to initialise the cache. Sorry for the noise!

Ralith:
Caching is now expanded to all the places where it's obviously applicable. Remaining work is benchmarking to see if this was actually a good idea.

Ralith:
Fiddled with inlining the new paths. LLVM really didn't want to do it, and it seems it was right -- inlining all the way down to the cache look-up had little or slightly negative effect. More profiling needed.

Ralith:
iterate_100k hot loop without caching (`--no-default-features`):

```
        │     pos.0 += vel.0;
   1.11 │ a0:   vmovss (%rdx,%rdi,4),%xmm0
  34.76 │       vaddss (%rsi,%rdi,4),%xmm0,%xmm0
  32.38 │       vmovss %xmm0,(%rsi,%rdi,4)
        │     self.position += 1;
   0.01 │       inc    %rdi
        │     if self.position == self.len {
   0.00 │ b2:   cmp    %rax,%rdi
  31.61 │     ↑ jne    a0
   0.02 │       mov    0x48(%rbx),%r13
```
with caching:
```
        │     ($($name::get($name, n),)*)
   0.19 │110:   mov        0x38(%rsp),%r10
  19.19 │       mov        0x40(%rsp),%r11
        │     self.position += 1;
   0.01 │11a:   lea        0x1(%rcx),%rdx
  18.59 │       mov        %rdx,0x48(%rsp)
        │     pos.0 += vel.0;
   2.06 │       vmovss     (%r11,%rcx,4),%xmm0
  22.68 │       vaddss     (%r10,%rcx,4),%xmm0,%xmm0
  19.29 │       vmovss     %xmm0,(%r10,%rcx,4)
   0.01 │       mov        %rdx,%rcx
        │     if self.position == self.len {
        │       cmp        %rax,%rcx
  17.90 │     ↑ jne        110
```
Note the extra costly `mov`s. Note also that the hot loop does *not* overlap with the code paths that actually change. I'm pretty well convinced this is LLVM's optimizations being fragile in non-local ways. While I'd like to fix that, I'm not confident that it can be fixed robustly, and I don't think it makes a meaningful difference outside of microbenchmarks. I'm inclined to merge this PR, but I'll let it sit for a bit longer in case anyone else has ideas/thoughts.

PR #401, opened 2025-05-17
Suppress spurious MIRI warnings

Ralith:


.github/workflows/rust.yml
```diff
@@ -59,6 +59,10 @@ jobs:
 
   miri:
     runs-on: ubuntu-latest
+    env:
+      # We only use permissive provenance ops for creating aligned
+      # dangling pointers, and strict provenance is a big MSRV bump.
```
adamreichold, line 64: You could copy that one polyfill from `sptr`, e.g.

```rust
fn invalid_ptr(addr: usize) -> *mut u8 {
    unsafe {
        #[allow(clippy::useless_transmute)]
        transmute(addr)
    }
}
```

PR #406, opened 2025-07-03
Added concurrent writers.

dragostis:
As long as types are different, you can now have multiple concurrent writers in a `ColumnBatch`.

Unfortunately, I don't think there's a way to add this without breaking the API.

### DIFF
--- a/src/batch.rs
+++ b/src/batch.rs
@@ -1,5 +1,5 @@
 use crate::alloc::collections::BinaryHeap;
-use core::{any::TypeId, fmt, mem::MaybeUninit, slice};
+use core::{any::TypeId, cell::Cell, fmt, mem::MaybeUninit, slice};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


dragostis:
The use case is data that's stored as an array of rows instead of an array of columns. Instead of having to iterate multiple times through the same data, you can do it in one go.

```rust
for (x, y, z) in [(bool, u32, f64); 1_024] {
    wx.push(x);
    wy.push(y);
    wz.push(z);
}
```

dragostis:
The example I gave was for row-major, but the actual need that we have are columns that generate intermediate data and need to be stored themselves. Having this API lets you get away by processing the column only once.

But even in the case of row-major: you might have an iterator API or very large files stored in row-major where changing is not really an option.

Ralith:
On second thought, `Mutex` is troublesome since it requires std. Happy to merge this as-is if you rebase it.

PR #421, opened 2025-12-26
impled query_shared for Entity

sanbox-irl:


src/query.rs
```diff
@@ -135,6 +135,7 @@ impl<T: Component> Query for &'_ T {
 }
 
 unsafe impl<T> QueryShared for &'_ T {}
+unsafe impl QueryShared for Entity {}
```
Ralith: style: group trait implementations under either the definition (and impl block, where applicable) of the type, or under the definition of the trait (as done for `impl Query for Entity`).

Ralith (changes requested):

src/query.rs
sanbox-irl: 👍

Ralith (approved):

PR #422, opened 2026-01-02
Fix `impl Send for Ref<T>`: use `Send` bound, instead of `Sync`

Philipp-M:
I don't think using `Sync` is the right bound here?

### DIFF
--- a/src/entity_ref.rs
+++ b/src/entity_ref.rs
@@ -136,7 +136,7 @@ impl<'a, T: Component> Ref<'a, T> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


adamreichold:
> I don't think using `Sync` is the right bound here?

Note that `Ref<'_, T>` is essentially `&T`, i.e. a shared reference to `T`, not `T` itself.

As you can read in the docs of [`Sync`](https://doc.rust-lang.org/stable/std/marker/trait.Sync.html), `&T` is `Send` if and only if `T` is `Sync`. So the code seems correct to me.

PR #436, opened 2026-03-04
Added an Iterator::fold implementation for QueryIter.

dragostis:
This implementation enables internal iteration for many Iterator methods that are implemented using `Iterator::fold` which greatly improves their performance.

Unfortunately, `Iterator::try_fold` cannot be implemented instead until `Try` becomes stable.

Improvements for the `for_each_100k` benchmark:

```
for_each 100k           time:   [32.154 µs 32.188 µs 32.221 µs]
                        change: [−79.807% −79.750% −79.704%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -836,6 +836,32 @@ impl<'q, Q: Query> Iterator for QueryIter<'q, Q> {
         let n = self.len();
         (n, Some(n))
     }
+
+    // We're implementing this because some `Iterator`-consuming methods are
+    // implemented using `Iterator::fold`. With an optmized version implemented
+    // without actually calling `Iterator::next`, they all benefit from
+    // improved performance. See `for_each_100k` bench.
+    //
+    // TODO: Once `Try` is stable, implement `Iterator::try_fold` instead. This
+    // will make *all* `Iterator`-consuming methods rely on that
+    // implementation.
+    fn fold<B, F>(mut self, mut init: B, mut f: F) -> B
+    where
+        Self: Sized,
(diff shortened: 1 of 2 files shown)


src/query.rs
```diff
@@ -836,6 +836,25 @@ impl<'q, Q: Query> Iterator for QueryIter<'q, Q> {
         let n = self.len();
         (n, Some(n))
     }
+
+    // TODO: once Try is stable, implement try_fold instead
+    fn fold<B, F>(mut self, mut init: B, mut f: F) -> B
```
Ralith, line 848: Include a comment documenting that this is motivated by better codegen for some cases, and citing the benchmark as evidence? That way we won't forget that this is important, and can easily confirm whether it still is.

Ralith (approved):
Amazing! I knew the regular iterator loop was getting pretty crap code-gen for simple bodies, but hadn't imagined there was such a simple work-around. I wonder why LLVM has an easier time with this form.

dragostis:
I also found out about this recently: all non-diverging `Iterator` consumers, e.g. `for_each`, `sum`, are implemented using `Iterator::fold` which defaults to using `Iterator::next`, but can be implemented in a more efficient way. It basically switches iteration from being pull to being push which is easy to optimize:

```rust
for arch in archetypes {
    // The loop below is very easy for LLVM to optimize compare to the pull version.
    for entity in arch {
        acc = fold(acc, entity);
    }
}
```

Issue #443, opened 2026-07-06
Add World::query_unchecked


da-shalev:
My systems run under a static scheduler, which I'm planning to open source: every system's component accesses are known ahead of time, and only provably disjoint systems ever run together. `query()'s` dynamic borrow checks can therefore never fail.

Wanted to know what you think about adding `World::query_unchecked`, so a query can skip the dynamic borrow checks when the caller already guarantees exclusivity.

```rs
/// Like `query_mut`, but unchecked.
/// # Safety
/// Caller must guarantee exclusive access to the archetypes `Q` matches.
pub unsafe fn query_unchecked<Q: Query>(&self) -> QueryMut<'_, Q>;
```

I realize that this is pretty niche. Draft PR: #444

Ralith:
I think this request is reasonable in principle. I'd like to see it motivated by meaningful empirical performance benefits in a realistic application.

da-shalev:
Okay, will experiment for fun sometime shortly in the future. Using my custom schedular.

PR #444, opened 2026-07-06
Add World::query_unchecked

da-shalev:
Issue: #443 

This adds:

```rust
pub unsafe fn query_unchecked<Q: Query>(&self) -> QueryMut<'_, Q>
```

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -873,12 +873,12 @@ impl<Q: Query> ExactSizeIterator for QueryIter<'_, Q> {
 
 /// A query builder that's convertible directly into an iterator
 pub struct QueryMut<'q, Q: Query> {
-    world: &'q mut World,
+    world: &'q World,
     _marker: PhantomData<fn() -> Q>,
 }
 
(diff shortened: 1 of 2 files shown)


src/query.rs
```diff
@@ -873,12 +873,12 @@ impl<Q: Query> ExactSizeIterator for QueryIter<'_, Q> {
...
 /// A query builder that's convertible directly into an iterator
 pub struct QueryMut<'q, Q: Query> {
-    world: &'q mut World,
+    world: &'q World,
     _marker: PhantomData<fn() -> Q>,
 }
 
 impl<'q, Q: Query> QueryMut<'q, Q> {
-    pub(crate) fn new(world: &'q mut World) -> Self {
+    pub(crate) fn new(world: &'q World) -> Self {
```
Ralith, line 881: For ease of maintenance, this should be `unsafe`, with the same invariant as on the `World` method.

Ralith (commented):
Implementation strategy LGTM; approval pending evidence that there's meaningful benefit.

PR #445, opened 2026-07-10
Implement Send and Sync for PreparedQuery

da-shalev:
`PreparedQuery` does not automatically implement `Send` and `Sync` because it retains pointer-capable fetch caches. Its prepared iterators and views already implement both traits when every `Q::Item<'a>` is `Send`; this applies the same bound to `PreparedQuery` itself.

`ChangeTracker` contains prepared queries, so it inherits `Send + Sync` transitively. This lets multithreaded schedulers store a `ChangeTracker` without introducing a local unsafe newtype.

The compile-time test covers prepared shared and mutable queries and confirms that `ChangeTracker` inherits both traits.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -1171,6 +1171,14 @@ pub struct PreparedQuery<Q: Query> {
     fetch: Box<[Option<Q::Fetch>]>,
 }
 
+// Safety: `Fetch::State` is `Send`, and cached fetches encode access to query
+// items, which this bound permits transferring between threads.
+unsafe impl<Q: Query> Send for PreparedQuery<Q> where for<'a> Q::Item<'a>: Send {}
+
+// Safety: `Fetch::State` is `Sync`, and cached fetches cannot be accessed
+// through a shared reference because every such method requires `&mut self`.
+unsafe impl<Q: Query> Sync for PreparedQuery<Q> {}
+
 impl<Q: Query> Default for PreparedQuery<Q> {
     fn default() -> Self {
         Self::new()


Ralith:
Thanks, this is an oversight. In fact, I think these impls should probably be on `PreparedQuery`... do you see any scenario where that would be a problem?

While fixes to `ChangeTracker` and `PreparedQuery` are absolutely welcome, be aware that you don't need either of them; the former is implemented in terms of the public API, and the latter should almost never be faster than the now built-in query caching.

da-shalev:
> Thanks, this is an oversight. In fact, I think these impls should probably be on `PreparedQuery`... do you see any scenario where that would be a problem?
> 
> While fixes to `ChangeTracker` and `PreparedQuery` are absolutely welcome, be aware that you don't need either of them; the former is implemented in terms of the public API, and the latter should almost never be faster than the now built-in query caching.

You're right, I'm updating the PR to have the Send + Sync impl's directly on PreparedQuery.  Having it on the ChangeTracker was the only thing I was concerned about because as you said PreparedQuery already caches itself.

Yes, thanks! I've looked into a event based impl of ChangeTracker.

The only issue I've ran into with event-based tracking is live change detection, so I explored using the provided ChangeTracker as a whole before over-optimizing with my own approach.

I'm exploring multiple forms of change detection before committing to one specific approach. And that's how I realized the oversight.

src/query.rs
```diff
@@ -1171,6 +1171,9 @@ pub struct PreparedQuery<Q: Query> {
     fetch: Box<[Option<Q::Fetch>]>,
 }
 
+unsafe impl<Q: Query> Send for PreparedQuery<Q> where for<'a> Q::Item<'a>: Send {}
+unsafe impl<Q: Query> Sync for PreparedQuery<Q> where for<'a> Q::Item<'a>: Send {}
```
Ralith: Are you sure the bound on `Sync` is correct? Bear in mind that all public methods are `&mut`. Also please include a safety comment.

Ralith (changes requested):

Ralith:
> as you said PreparedQuery already caches itself.

To be clear: *all* queries are automatically cached. `PreparedQuery` is mostly redundant now.

src/query.rs
```diff
@@ -1171,6 +1171,14 @@ pub struct PreparedQuery<Q: Query> {
     fetch: Box<[Option<Q::Fetch>]>,
 }
 
+// Safety: `Fetch::State` is `Send`, and cached fetches encode access to query
+// items, which this bound permits transferring between threads.
+unsafe impl<Q: Query> Send for PreparedQuery<Q> where for<'a> Q::Item<'a>: Send {}
```
Ralith, line 1176: I'm actually not convinced even this bound is justified. There's no `Q::Item`s being sent, and all methods that might get you one require you to supply a `World`, which already entitles you to that.

Maybe `Fetch` should imply `Send` directly?

Issue #351, opened 2023-11-02
Less than ideal codegen for iteration


dragostis:
The way query iteration is currently implemented leads to inefficient code. On my machine, `iterate_mut_100k` runs in around 29us. However, running the same iteration through explicit archetypes leads to much better performance at around 10us. Looking at the assembly, the difference is loop unrolling and better auto-vectorization.

From previous experience writing code like this in Rust, I'd say that the issue is caused by having to deal with case where we need to move to the next archetype when iterating over the query. I haven't yet tested, but I'd expect similar or even worse performance were the archetypes `Iterator::chain`ed together.

Here is the code that runs ~2.9x faster:
```rust
fn iterate_mut_100k_archetypes(b: &mut Bencher) {
    let mut world = World::new();
    for i in 0..100_000 {
        world.spawn((Position(-(i as f32)), Velocity(i as f32)));
    }
    b.iter(|| {
        for archetype in world.archetypes() {
            if let (Some(mut pos), Some(vel)) = (archetype.get::<&mut Position>(), archetype.get::<&Velocity>()) {
                for (pos, vel) in pos.iter_mut().zip(vel.iter()) {
                    pos.0 += vel.0;
                }
            }
        }
    })
}
```

I'm opening this issue to start a discussion about how some of this performance could be tapped into without having to rely on explicitly iterating through all the archetypes.

Ralith:
I'm not overly concerned about this because I think the difference will disappear outside of trivial loop bodies (i.e. in real use), but improvements would certainly be welcome.

dragostis:
I mostly agree about real use, but there are many cases where this would be useful. For example, in [forma](https://github.com/google/forma) there's a decently complex rasterization pipeline that relies on rustc/LLVM auto-vectorizing every line. (of which there can be many)

I'll try to see if I can get LLVM to spit out some interesting optimization remarks that might be useful and thanks for the timely reply!

PR #455, opened 2026-08-20
Use range for instead of next in fold

Koranir:
Related (but doesn't fix): #351

Allows the compiler to optimise more by giving it extra information to work with, namely giving it proper length information instead of hoping it peeks through repeated while let calls.

By all rights it should be able to optimise without this change but possibly due to some non-optimal optimisation ordering it doesn't end up being able to vectorize loops, even though it inlines everything to essentially a range for in the end.

The for_each 100k benchmark goes from 100 µs to 22 µs on my system, about a 4x speedup (which tracks with using 128-bit SIMD instructions). Now this is a very contrived benchmark and I suspect actual gains will be much less impactful, but might as well..?

This doesn't change the speed of the iterate mut 100k benchmark, which doesn't use fold (perhaps there should be more documentation explaining that fold/for_each should be used when possible).

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -852,9 +852,7 @@ impl<'q, Q: Query> Iterator for QueryIter<'q, Q> {
         F: FnMut(B, Self::Item) -> B,
     {
         loop {
-            while let Some(components) = unsafe { self.iter.next(self.world.entities_meta()) } {
-                init = f(init, components);
-            }
+            init = unsafe { self.iter.fold(self.world.entities_meta(), init, &mut f) };
 
             if let Some(iter) = unsafe { self.archetypes.next(self.world) } {
                 self.iter = iter;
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Thanks!

Ralith:
> about a 4x speedup (which tracks with using 128-bit SIMD instructions).

Out of curiosity, do we actually see that in the output?

> perhaps there should be more documentation explaining that fold/for_each should be used when possible

Given that the overwhelming majority of applications won't benefit, I'm reluctant to broadly recommend a much less ergonomic pattern.

newDINO:
The for_each 100k benchmark has much higher improvement on my device: from 168.9us to 8.14us, which is around 95% percent time decrease !

The performance improvement is not only due to SIMD, but also the reduced branches. The next() based method has 2 branches per item while the for-range based only has 1 branch per item.

Hoping to see this published into crates.io soon.

newDINO:
> The performance improvement is not only due to SIMD, but also the reduced branches. The next() based method has 2 branches per item while the for-range based only has 1 branch per item.

My mistake, this improvement is actually due to #436

newDINO:
> The for_each 100k benchmark has much higher improvement on my device: from 168.9us to 8.14us, which is around 95% percent time decrease !

Another interesting point is that this result is tested when using `hecs` from another crate. When benchmarking in the `hecs` crate, the improvement is only 77%: from 36us to 8.4us.

Ralith:
> Hoping to see this published into crates.io soon.

> My mistake, this improvement is actually due to https://github.com/Ralith/hecs/pull/436

Note that #436 is already in the current release.

PR #456, opened 2026-08-31
Add fold method to Batch and PreparedQueryIter

newDINO:
Add their fold implementation and benchmarks.

For `PreparedQueryIter`, there is a −67% time decrease: from 25us to 8.1us.
For `Batch`, there is basically no improvement. But due to the uncertainty of compiler optimization, I think this may still be useful.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -175,6 +175,41 @@ fn for_each_100k(c: &mut Criterion) {
     });
 }
 
+fn for_each_prepared_100k(c: &mut Criterion) {
+    let mut world = World::new();
+    for i in 0..100_000 {
+        world.spawn((Position(-(i as f32)), Velocity(i as f32)));
+    }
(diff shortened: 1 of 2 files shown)


newDINO:
I also think it is worth mentioning that `for_each` has better performance than `for`-loop in the documentation, especially when the there is no much computation per iteration, and it can achieve performance close to iterating over rust built-in `slice`s.

`bevy_ecs` also has this kind of mentioning (https://docs.rs/bevy_ecs/latest/bevy_ecs/system/struct.Query.html#iteratorfor_each).

Ralith (approved):

Ralith:
Thanks! If you're currently using prepared queries, I recommend trying without them now that query caching is automatic.

newDINO:
Thanks for the recommendation. Currently I'm not using `PreparedQuery` due to its `!Sync` limitations and that it's not very ergonomic.

On my machine, the automatically cached query only takes tens of nanoseconds, which is highly efficient.

PR #458, opened 2026-09-02
Require manual Component implementation

Ralith:
Alternative to https://github.com/Ralith/hecs/pull/454. This is a major breaking change for which I am seeking feedback. It is motivated by several user reports of difficult-to-diagnose errors, and general agreement that inserting foreign types as components is rare and not especially important.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+# Unreleased
+
+### Changed
+
+- `Component` is no longer automatically implemented for all `Send + Sync + 'static` types. Types
+  must now implement it explicitly, either by hand (`impl Component for Position {}`) or with the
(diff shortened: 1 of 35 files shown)


sanbox-irl:
I implemented this in Fields of Mistria's engine and it + about a minute of adding derives to components, and we were done with no issues. It discovered no new bugs.

One minor issue however was components which had a type parameter(s). This isn't novel to this derive macro (clone has the same issue, for example), where:
```
#[derive(hecs::Component)]
pub struct Foo<T>(T);
```
only works when T: Component. That's the standard default and is fine, but it did require me to use my brain a bit to manually write a wrapper impl, which was the appropriate case here.

In any manner, it was a very successful test for us!

Ralith:
Yeah, sometimes folks offer helper attributes to customize generated bounds, but it's *so* trivial to implement the trait that I think the simplicity of doing so with conventional syntax wins out.

Issue #461, opened 2026-09-02
Cloning an empty or ZST-only `EntityBuilderClone` allocates a zero-size layout


DRMacIver:
```rust
use hecs::EntityBuilderClone;

fn main() {
    let empty = EntityBuilderClone::new();
    let _clone = empty.clone();
}
```

Under Miri:

```
error: Undefined Behavior: creating allocation with size 0
   --> src/entity_builder.rs:411:49
    |
411 |                 storage: NonNull::new_unchecked(alloc(self.layout)),
    |                                                 ^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
```

`GlobalAlloc::alloc` documents a zero-size layout as undefined behaviour, and `NonNull::new_unchecked` on a null return would be too. A builder holding only zero-sized components does the same; one holding anything else is fine. Nothing goes wrong outside Miri.

Tested on hecs 0.11.1 and current master (677a94e).

Issue #460, opened 2026-09-02
`EntityBuilderClone` returns the wrong component after a round trip through `BuiltEntityClone`


DRMacIver:
```rust
use hecs::EntityBuilderClone;

#[derive(Clone)]
struct Small(u8);
#[derive(Clone)]
struct Big(u64);

fn main() {
    let mut b = EntityBuilderClone::new();
    b.add(Small(7));
    b.add(Big(0x4242_4242_4242_4242));
    let rebuilt: EntityBuilderClone = b.build().into();
    println!("get::<&Small>() = {:?}", rebuilt.get::<&Small>().map(|s| s.0));
}
```

Output:

```
get::<&Small>() = Some(66)
```

Expected `Some(7)`. Adding `Big` before `Small` returns `Some(7)`, so it takes the smaller-aligned component going in first.

Calling `add` on the round-tripped builder reads out of bounds:

```rust
let mut b = EntityBuilderClone::new();
b.add(Small(7));
b.add(Big(0x4242_4242_4242_4242));
let mut rebuilt: EntityBuilderClone = b.build().into();
rebuilt.add(Small(9));
```

Under Miri:

```
error: Undefined Behavior: memory access failed: attempting to access 8 bytes, but got alloc1369 which is only 1 byte from the end of the allocation
   --> src/entity_builder.rs:351:17
    |
351 |                 ptr::copy_nonoverlapping(ptr, storage, ty.layout().size());
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
```

Without Miri that program spawns an entity with `Small == 7` and `Big == 265`.

Two components of the same size — two `i32` newtypes, say — give wrong values rather than an out-of-bounds read, depending on `TypeId` ordering.

Tested on hecs 0.11.1 and current master (677a94e).

Issue #459, opened 2026-09-02
`ColumnBatchBuilder::build` leaks the written components when it returns `Err`


DRMacIver:
```rust
use std::sync::Arc;
use hecs::ColumnBatchType;

fn main() {
    let mut ty = ColumnBatchType::new();
    ty.add::<Arc<()>>();
    ty.add::<u32>();

    let mut builder = ty.into_batch(2);
    let value = Arc::new(());
    {
        let mut writer = builder.writer::<Arc<()>>().unwrap();
        writer.push(value.clone()).unwrap();
        writer.push(value.clone()).unwrap();
    }

    // The u32 column was never written, so build() fails.
    println!("build() is_err = {}", builder.build().is_err());
    println!("strong_count = {}", Arc::strong_count(&value));
}
```

Output:

```
build() is_err = true
strong_count = 3
```

The two `Arc`s pushed into the batch are never dropped, so the count stays at 3 instead of returning to 1. Dropping an unbuilt builder without calling `build` releases them correctly — this only happens on the `Err` return from `build`.

Tested on hecs 0.11.1 and current master (677a94e).

Issue #450, opened 2026-07-27
ColumnBatchBuilder leaks its written components when dropped without a successful build()


DRMacIver:
Components pushed into a `ColumnBatchBuilder` are never dropped if the builder is dropped without being consumed by a successful `build()`. This program pushes two components, drops the builder, and the two values' destructors never run:

```rust
use std::sync::atomic::{AtomicI64, Ordering};
use hecs::ColumnBatchType;

static LIVE: AtomicI64 = AtomicI64::new(0);

struct Tracked(#[allow(dead_code)] String);
impl Tracked {
    fn new() -> Self {
        LIVE.fetch_add(1, Ordering::SeqCst);
        Tracked("hello".into())
    }
}
impl Drop for Tracked {
    fn drop(&mut self) {
        LIVE.fetch_sub(1, Ordering::SeqCst);
    }
}

fn main() {
    let mut ty = ColumnBatchType::new();
    ty.add::<Tracked>();
    let builder = ty.into_batch(3); // room for 3 entities
    {
        let mut w = builder.writer::<Tracked>().unwrap();
        w.push(Tracked::new()).ok().unwrap();
        w.push(Tracked::new()).ok().unwrap();
    }
    drop(builder); // only 2 of 3 rows written, so the builder is never built
    println!("live Tracked after dropping the builder: {}", LIVE.load(Ordering::SeqCst));
}
```

Output:

```
live Tracked after dropping the builder: 2
```

The two `Tracked` values are still live after the builder is dropped; their `Drop` never runs. Expected output is `0`.

The same happens when a partially-filled builder is finalized with `build()`, which returns `Err(BatchIncomplete)` and leaks whatever was written:

```rust
    let builder = ty.into_batch(3);
    {
        let mut w = builder.writer::<Tracked>().unwrap();
        w.push(Tracked::new()).ok().unwrap();
        w.push(Tracked::new()).ok().unwrap();
    }
    let result = builder.build(); // Err(BatchIncomplete): one row short
    assert!(result.is_err());
    println!("live Tracked after build() returned Err: {}", LIVE.load(Ordering::SeqCst));
```

Output:

```
live Tracked after build() returned Err: 2
```

A fully-filled builder that is consumed by a successful `build()` and spawned into a `World` drops its components correctly; the leak only affects the drop-without-build and the `Err(BatchIncomplete)` paths.

[...]

PR #462, opened 2026-09-02
Add hegel property tests

DRMacIver:
Following up on #450, where you said you'd be happy to review these for inclusion. Claude Code wrote the tests and this description. Please read this as an offer of code rather than an imposition: if it is more than you want to own, or the wrong shape for the repo, say so and no offence will be taken.

Three tests are `#[ignore]`d, one each for #459, #460 and #461, which came out of this suite. The inputs that trigger those three are guarded rather than left to fail, with a comment saying why, so the suite stays clean under `cargo miri test --all-features --lib --tests`.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -10,7 +10,7 @@ use std::cell::Cell;
 use std::collections::{BTreeMap, HashSet};
 
 use hecs::{Entity, EntityBuilder, World};
-use hegel::generators as gs;
+use hegel::generators::{self as gs, Generator};
 use serde::{Deserialize, Deserializer, Serialize};
 
 /// Component universe: two same-shaped payload components, a zero-sized marker,
@@ -99,17 +99,10 @@ pub fn val() -> impl gs::PrintableGenerator<i32> {
     gs::integers::<i32>().min_value(-3).max_value(3)
 }
 
-/// Draw from `pool`, which deliberately retains despawned handles.
-pub fn pick(tc: &hegel::TestCase, pool: &[Entity]) -> Option<Entity> {
-    if pool.is_empty() {
-        return None;
-    }
-    let i = tc.draw(
-        gs::integers::<usize>()
-            .min_value(0)
-            .max_value(pool.len() - 1),
-    );
-    Some(pool[i])
+/// A handle from `pool`, which deliberately retains despawned handles. `pool`
+/// must not be empty.
+pub fn pick(pool: &[Entity]) -> impl gs::PrintableGenerator<Entity> + '_ {
+    gs::sampled_from(pool).print_as_debug()
 }
 
 /// An arbitrary subset of the component universe, as drawn payloads. `D`
@@ -142,6 +135,12 @@ pub fn specs(tc: &hegel::TestCase) -> Spec {
         c: tc.draw(gs::booleans()),
         d: tc.draw(gs::optional(val())),
     }
+}
+
+/// `specs()` without `D`, which is not `Clone` and so cannot go into an
+/// `EntityBuilderClone`.
+pub fn specs_without_d() -> impl gs::PrintableGenerator<Spec> {
+    specs().map(|s| Spec { d: None, ..s })
 }
 
 pub fn make_builder(s: Spec) -> EntityBuilder {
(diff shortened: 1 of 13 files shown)


tests/common/mod.rs
```diff
@@ -0,0 +1,266 @@
+//! Shared fixtures for the property tests: a fixed component universe, an
+//! observational fingerprint of a `World`, and twin worlds replayed from one
+//! drawn history.
+//!
+//! Each integration-test binary compiles this module separately, so parts of it
+//! are unused in some binaries.
```
Ralith: Moving this into a helper crate would avoid redundant compilation and ensure we get dead code detection for non-public items.

tests/common/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+/// Component universe: two same-shaped payload components, a zero-sized marker,
+/// and a drop-tracked non-`Copy` component.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct A(pub i32);
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct B(pub i32);
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct C;
+
+thread_local! { static D_LIVE: Cell<i64> = const { Cell::new(0) }; }
```
Ralith: Is global state absolutely necessary? Normally I do liveness tests using an `Arc`. Even when deserializing, `DeserializeSeed` can be used. Then we wouldn't need to worry about `assert_d_balanced_at_start`.

tests/common/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+    assert_eq!(d_live(), 0, "live D count nonzero at case start");
+}
+
+/// Under Miri each operation is interpreted, so these tests run a much smaller
+/// search and serve as a UB oracle for hecs's unsafe component moves rather
+/// than as a search for logic bugs. Two consequences: hegel's `TooSlow` check
+/// would report on interpreter speed rather than on anything about the tests,
+/// and Miri's isolation denies the file and random-device access hegel needs
+/// for its failure database and its seed, so both are turned off.
+#[cfg(miri)]
```
Ralith: nit: prefer to put conditional compilation *inside* the function, so there's no risk of the signatures/names drifting.

tests/common/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+/// Worlds are kept small so that handle collisions, empty archetypes and
+/// stale handles all occur often; the bound protects the tests' runtime, not
+/// any hecs contract.
+#[cfg(miri)]
+pub const MAX_ENTITIES: u32 = 4;
+#[cfg(not(miri))]
+pub const MAX_ENTITIES: u32 = 8;
+
+/// Component payloads are small so that distinct values recur across entities,
+/// which is what makes a swapped or stale component visible.
```
Ralith: Wouldn't reusing values make those cases *less* visible?

tests/common/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+pub fn val() -> impl gs::PrintableGenerator<i32> {
+    gs::integers::<i32>().min_value(-3).max_value(3)
+}
+
+/// Draw from `pool`, which deliberately retains despawned handles.
```
Ralith: This documentation and/or name are unclear. Why does an `&[Entity]` necessarily have anything to do with spawned-ness in some world? Is this not a reimplementation of `sampled_from`?

tests/common/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+        gs::integers::<usize>()
+            .min_value(0)
+            .max_value(pool.len() - 1),
+    );
+    Some(pool[i])
+}
+
+/// An arbitrary subset of the component universe, as drawn payloads. `D`
+/// instances are materialized only when a bundle is built, so the drop count
+/// tracks exactly the instances that entered a world.
```
Ralith: What's a component universe, how does this represent a subset of one, and why is that called a "Spec"? What does any of this have to do with bundles or drop count?

From the method, it sounds like this struct describes an *entity*, but nothing in the doc comment seems to support that.

tests/common/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+pub fn specs(tc: &hegel::TestCase) -> Spec {
+    Spec {
+        a: tc.draw(gs::optional(val())),
+        b: tc.draw(gs::optional(val())),
+        c: tc.draw(gs::booleans()),
+        d: tc.draw(gs::optional(val())),
+    }
+}
+
+pub fn make_builder(s: Spec) -> EntityBuilder {
```
Ralith: Should this be a method on `Spec`?

tests/common/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+        total,
+        world.len(),
+        "{label}: archetype lengths sum to {total}, world.len() is {}",
+        world.len()
+    );
+}
+
+/// Build `n_worlds` observationally identical worlds by replaying one drawn
+/// history (spawns of arbitrary specs, then a drawn subset of despawns) into
+/// each, and return them with a target pool that retains the despawned handles.
```
Ralith: What's a target pool? Would it be clearer to just say "`Vec`"?

Does "retains" mean it includes *just* the despawned handles, or all entity handles that were ever spawned?

tests/common/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+///
+/// `World` is not `Clone`, so this is how a relation between two executions
+/// "from the same state" is set up. It relies on hecs allocating handles
```
Ralith: It's not difficult to duplicate a world containing known components if you want to; there's an example for that. Would that be simpler?

Ralith (changes requested):
I found the documentation in the first commit quite opaque. Before I review further, can you confirm that you've manually reviewed all AI output in this PR and edited it for clarity?

DRMacIver:
@Ralith Thanks for the detailed review! I'll try to get your comments addressed shortly

> I found the documentation in the first commit quite opaque. Before I review further, can you confirm that you've manually reviewed all AI output in this PR and edited it for clarity?

Sorry about that.

FWIW I had manually reviewed all AI output the first time (though I hadn't made any edits myself), but possibly not sufficiently carefully. I think it's more that we're familiar with different areas (I know Hegel very well and hecs not at all, and presumably vice versa) so the things that are confusing didn't leap out at me. Also TBH I've probably gotten a little too used to reading Claude output. I'll do a much more careful human review for clarity before I ask for a re-review.

Ralith:
> Sorry about that.

No worries. I really appreciate all the issues this work is finding!

> I think it's more that we're familiar with different areas (I know Hegel very well and hecs not at all, and presumably vice versa) so the things that are confusing didn't leap out at me.

Totally willing to believe this! The target audience of these comments is probably also not hegel experts, though, and I had trouble finding definitions of e.g. "universe" in the hegel docs.

DRMacIver:
> The target audience of these comments is probably also not hegel experts, though, and I had trouble finding definitions of e.g. "universe" in the hegel docs.

Yeah, I agree on further reading that the comments were just bad. I think I didn't pay enough attention to them on first read and was more focused on the code.

DRMacIver:
@Ralith The more I dig into this, the more room for improvement I find on this PR. I'm going to close it, rework it heavily, and will reopen a new one when I'm satisfied. Sorry once again for opening this PR before it was up to standard. Mea culpa.

PR #470, opened 2026-09-08
Changed DrainOnDrop to use optimized fold.

dragostis:
In order for auto-vectorizer to kick in when using the ChangeTracker, the DrainOnDrop iterator needs to forward the fold implementation apart from implement next.

### DIFF
--- a/src/change_tracker.rs
+++ b/src/change_tracker.rs
@@ -82,7 +82,7 @@ where
     pub fn added(&mut self) -> impl ExactSizeIterator<Item = (Entity, &T)> + '_ {
         self.tracker.added_components.clear();
         self.added = true;
-        DrainOnDrop(
+        DrainOnDrop::new(
             self.tracker
                 .added
                 .query_mut(self.world)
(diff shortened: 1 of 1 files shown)


src/change_tracker.rs
```diff
@@ -148,25 +148,45 @@ where
...
     fn len(&self) -> usize {
-        self.0.len()
+        self.0.as_ref().map_or(0, ExactSizeIterator::len)
     }
 }
 
 impl<T: Iterator> Drop for DrainOnDrop<T> {
     fn drop(&mut self) {
-        for _ in &mut self.0 {}
+        if let Some(iter) = self.0.take() {
```
Ralith, line 187: We should be able to avoid introducing an `Option` here by borrowing the iterator using [the implementation of `Iterator` for `&mut I`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#impl-Iterator-for-%26mut+I)

Ralith (changes requested):

src/change_tracker.rs
dragostis: Calling `fold` on `&mut self.0` doesn't route it through the optimized `fold` as far as I can tell. `Iterator for &mut I` implements its `fold` through `I::try_fold` which is not overridden and ends up relying on `I::next`.

Ralith: Oh, that's annoying. Leave another TODO for when Try stabilizes, I guess?

PR #471, opened 2026-09-09
Fix double-drop when a component's Drop panics

tooson9010-spec:
Found while researching panic-safety in Rust containers.

## Summary

Three places destroy or move components before committing the state that says
which slots are live. `Component` has no bound excluding a panicking `Drop`, so
a caller only needs to catch the unwind and touch the container again. All three
are double-drops (CWE-415) reachable from safe Rust.

| Location | Skipped commit | Second drop |
| --- | --- | --- |
| `Archetype::clear` | `self.len = 0` | `Drop for Archetype` calls `clear` again |
| `Archetype::remove` | `self.len = last` | the slot already dropped, and the duplicate left by `copy_nonoverlapping` |
| `CommandBuffer::run_on` | `self.components.clear()` | `CommandBuffer::clear` drops components already moved into the world |

`run_on` already carries a comment saying "Wipe out component references so
`clear` doesn't try to double-free". It just does not run when a command
unwinds.

## Fix

`clear` takes the length before running any destructor. Components not reached
are leaked, which is safe.

`remove` runs the loop under a guard. On unwind the guard finishes the backfill
for the remaining types without running further destructors, moves the entity
id, and commits the length. The dropped slot is overwritten by the moved entity
and the duplicate falls out of range, so the archetype is left consistent.
Components of the removed entity past the panic are leaked.

`run_on` tracks how many components the world has taken and splits the two
groups in a guard. Those already handed over are dropped from the list without
running destructors, the rest are destroyed normally. No leak on this path.

`World::despawn` is not covered. It updates its entity index from the value
`remove` returns, and that is skipped when the unwind passes through, so the
moved entity keeps a stale index entry. That is a logical inconsistency, not
memory unsafety. I left it alone since the right shape looks like your call.

## Verification

Three regression tests count destructor invocations. Against the unpatched
code `clear` runs 7 destructors where 4 components exist, `run_on` runs 3
where 2 exist, and `despawn` runs 8 where 6 exist. With the fix `clear`
destroys 3 of the 4 and `run_on` 2 of 2; the `despawn` test asserts only
that the count stays at or below 6, since how far the loop gets before the
unwind depends on the order the archetype stores its types in.

The double-frees were confirmed earlier with AddressSanitizer (`attempting
double-free`) on the unpatched code, through `World::clear`,
`World::despawn`, `World::spawn_at` and `World::spawn_column_batch_at`.
Existing tests pass and `cargo build --no-default-features` still builds.
Confirmed on 0.11.1.

### DIFF
--- a/src/archetype.rs
+++ b/src/archetype.rs
@@ -66,15 +66,20 @@ impl Archetype {
     }
 
     pub(crate) fn clear(&mut self) {
+        // Take the length before running any destructor. `Drop for Archetype`
+        // calls this function, so if a component's `Drop` unwinds and the
+        // length is still set, every component is destroyed a second time.
+        // Components not reached before the unwind are leaked instead.
+        let len = self.len;
+        self.len = 0;
         for (ty, data) in self.types.iter().zip(&*self.data) {
-            for index in 0..self.len {
+            for index in 0..len {
                 unsafe {
                     let removed = data.storage.as_ptr().add(index as usize * ty.layout.size());
                     (ty.drop)(removed);
                 }
             }
         }
-        self.len = 0;
     }
 
     /// Whether this archetype contains `T` components
--- a/src/command_buffer.rs
+++ b/src/command_buffer.rs
@@ -143,10 +143,33 @@ impl CommandBuffer {
 
     /// Run recorded commands on `world`, clearing the command buffer
     pub fn run_on(&mut self, world: &mut World) {
-        for i in 0..self.cmds.len() {
-            match mem::replace(&mut self.cmds[i], Cmd::Despawn(Entity::DANGLING)) {
+        // Components handed to `world` must not be destroyed by `clear`, so the
+        // count of consumed components is committed as the loop goes. Without
+        // this a panicking command leaves them owned twice.
+        struct Reset<'a> {
+            buffer: &'a mut CommandBuffer,
+            consumed: usize,
+        }
+
+        impl Drop for Reset<'_> {
+            fn drop(&mut self) {
+                // Already moved into the world: forget them here.
+                self.buffer.components.drain(..self.consumed);
+                // Still owned by the buffer: destroy them normally.
+                self.buffer.clear();
+            }
+        }
+
+        let mut state = Reset {
+            buffer: self,
+            consumed: 0,
+        };
+
+        for i in 0..state.buffer.cmds.len() {
+            match mem::replace(&mut state.buffer.cmds[i], Cmd::Despawn(Entity::DANGLING)) {
                 Cmd::SpawnOrInsert(entity) => {
-                    let components = self.build(entity.components);
+                    let end = entity.components.end;
+                    let components = state.buffer.build(entity.components);
                     match entity.entity {
                         Some(entity) => {
                             // If `entity` no longer exists, quietly drop the components.
@@ -156,6 +179,7 @@ impl CommandBuffer {
                             world.spawn(components);
                         }
                     }
+                    state.consumed = end;
                 }
                 Cmd::Remove(remove) => {
                     (remove.remove)(world, remove.entity);
@@ -168,10 +192,6 @@ impl CommandBuffer {
                 }
             }
         }
-        // Wipe out component references so `clear` doesn't try to double-free
-        self.components.clear();
-
-        self.clear();
     }
 
     fn build(&mut self, components: Range<usize>) -> RecordedEntity<'_> {
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -788,6 +788,43 @@ fn clear() {
 }
 
 #[test]
+fn clear_does_not_double_drop_when_a_destructor_panics() {
+    use std::panic::{catch_unwind, AssertUnwindSafe};
+    use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+
+    static DROPS: AtomicUsize = AtomicUsize::new(0);
+    static ARMED: AtomicBool = AtomicBool::new(true);
+
+    struct Bomb(usize);
+
+    impl Drop for Bomb {
+        fn drop(&mut self) {
+            DROPS.fetch_add(1, Ordering::SeqCst);
+            // Fire once. A second panic while unwinding would abort.
+            if self.0 == 2 && ARMED.swap(false, Ordering::SeqCst) {
+                panic!("boom");
+            }
+        }
+    }
+
+    let mut world = World::new();
+    for i in 0..4 {
+        world.spawn((Bomb(i),));
+    }
+
+    let unwound = catch_unwind(AssertUnwindSafe(|| world.clear()));
+    assert!(unwound.is_err());
+
+    drop(world);
+
+    assert_eq!(
+        DROPS.load(Ordering::SeqCst),
+        3,
+        "components destroyed before the unwind must not be destroyed again"
+    );
+}
+
+#[test]
 fn remove_missing() {
     let mut world = World::new();
     let e = world.spawn(("abc", 123));
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -825,6 +825,44 @@ fn clear_does_not_double_drop_when_a_destructor_panics() {
 }
 
 #[test]
+fn command_buffer_does_not_double_drop_when_a_command_panics() {
+    use std::panic::{catch_unwind, AssertUnwindSafe};
+    use std::sync::atomic::{AtomicUsize, Ordering};
+
+    static DROPS: AtomicUsize = AtomicUsize::new(0);
+
+    struct Tracked;
+
+    impl Drop for Tracked {
+        fn drop(&mut self) {
+            DROPS.fetch_add(1, Ordering::SeqCst);
+        }
+    }
+
+    let mut world = World::new();
+    let mut buffer = CommandBuffer::new();
+
+    // Handed to the world by the first command.
+    buffer.spawn((Tracked,));
+    // Unwinds before run_on can wipe the component list.
+    buffer.queue(|_| panic!("boom"));
+    // Never reached; still owned by the buffer.
+    buffer.spawn((Tracked,));
+
+    let unwound = catch_unwind(AssertUnwindSafe(|| buffer.run_on(&mut world)));
+    assert!(unwound.is_err());
+
+    drop(buffer);
+    drop(world);
+
+    assert_eq!(
+        DROPS.load(Ordering::SeqCst),
+        2,
+        "a component moved into the world must not be destroyed by the buffer"
+    );
+}
+
+#[test]
 fn remove_missing() {
     let mut world = World::new();
     let e = world.spawn(("abc", 123));


tests/tests.rs
```diff
@@ -787,6 +787,43 @@ fn clear() {
     assert_eq!(world.iter().count(), 0);
 }
 
+#[test]
+fn clear_does_not_double_drop_when_a_destructor_panics() {
+    use std::panic::{catch_unwind, AssertUnwindSafe};
+    use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+
+    static DROPS: AtomicUsize = AtomicUsize::new(0);
+    static ARMED: AtomicBool = AtomicBool::new(true);
```
Ralith: Could we use `Arc` rather than globals for this sort of state in this and following tests? That would make them easier to reason about.

tests/tests.rs
```diff
@@ -787,6 +787,43 @@ fn clear() {
...
+
+    struct Bomb(usize);
+
+    impl Drop for Bomb {
+        fn drop(&mut self) {
+            DROPS.fetch_add(1, Ordering::SeqCst);
```
Ralith: Why do we need `SeqCst` when there's only one thread accessing these? Would `Relaxed` be sufficient?

src/archetype.rs
```diff
@@ -324,19 +324,68 @@ impl Archetype {
...
-            let removed = data.storage.as_ptr().add(index as usize * ty.layout.size());
-            if drop {
-                (ty.drop)(removed);
+
+        // Each iteration destroys the component at `index` and then backfills
+        // it from `last`, leaving a bitwise duplicate behind until `self.len`
+        // is committed. If a destructor unwinds partway through, the guard
+        // finishes the backfill for the remaining types without running any
+        // further destructors, then commits the length. Components of the
+        // removed entity that were not reached are leaked, which is safe.
```
Ralith: This is a lot of extra, somewhat repetitive code. Would it be simpler to wrap `catch_unwind` around the existing `drop` call, then resume unwinding after the loop finishes? That could even allow us to continue dropping components after one panics on drop.

Ralith (changes requested):
Thanks!

tests/tests.rs
tooson9010-spec: Thanks for the review.

`DropLog` holds the count and the fire-once flag, and each component keeps an `Arc` to it.

tests/tests.rs
tooson9010-spec: Thanks. Switched to `Relaxed`.

src/archetype.rs
tooson9010-spec: Agreed, that would be simpler.

`catch_unwind` needs `std` though, and `remove` is on the `no_std` path too.
So it'd be `#[cfg(feature = "std")]` for the `catch_unwind` version and
something like the current guard as the fallback, which trades the repetition
for two code paths. Or std-only, leaving `no_std` unfixed. Which would you
prefer?

Ralith: Hmm, better not break no-std, and having two versions would defeat the point. Maybe extract a helper function with `drop: bool` or something?

tooson9010-spec: Done. `remove_one` takes the `drop` flag, and the guard passes `false` so
unwinding doesn't run another destructor. Please let me know if you meant a
different split.

Running Miri locally, the `Arc` rewrite trips its leak check. The fix leaks
the components a panic didn't reach, and each of those holds an `Arc`, so
`cargo miri test --lib --tests` reports two leaked allocations. The tests
themselves all pass.

I can add `#[cfg_attr(miri, ignore)]` to the three tests, or drop the `Arc`
again, whichever you prefer.

Ralith: Oh, miri treats leaks as errors? That's annoying...

Having miri coverage for all unsafe code is important. I guess we can compromise on thread-locals rather than statics.

tooson9010-spec: Thanks, that works. `cargo miri test --all-features --lib --tests` is clean
now, the three panic-safety tests included.

src/archetype.rs
```diff
@@ -319,19 +324,57 @@ impl Archetype {
     /// Returns the ID of the entity moved into `index`, if any
     pub(crate) unsafe fn remove(&mut self, index: u32, drop: bool) -> Option<u32> {
         let last = self.len - 1;
-        for (ty, data) in self.types.iter().zip(&*self.data) {
-            let removed = data.storage.as_ptr().add(index as usize * ty.layout.size());
-            if drop {
-                (ty.drop)(removed);
-            }
-            if index != last {
```
adamreichold, line 327: Wouldn't it be simpler to first move all components, update the length and then finally always drop the last (or one-past-last by then) component values when everything is committed?

adamreichold: Or rather, always do all the copying in the `Guard` destructor and only the actual `Drop`ing in the straight-line code path?

Ralith:
> Each test runs on its own thread, so a thread-local gives the same isolation without allocating.

Is this still true when passing `--test-threads=1`, or do we need to explicitly reinitialize the thread-locals in every test that uses them?

Ralith:
Also, for easier review, please squash changes from feedback into the commits that were changed.

src/archetype.rs
tooson9010-spec: Sorry for the slow reply. 
Done. the guard does all the copying now and the
straight-line path only runs the destructors. The `progress` counter and the
`remove_one` helper are gone.

tooson9010-spec:
Sorry again for the slow reply.

You're right, it isn't. The tests reset the counters at the start now, and the
suite passes with `--test-threads=1`.

Also squashed the feedback commits into the ones they changed.

src/archetype.rs
```diff
@@ -319,19 +324,51 @@ impl Archetype {
...
+            last,
+        };
+        if drop {
+            for (ty, data) in guard.archetype.types.iter().zip(&*guard.archetype.data) {
+                let removed = data.storage.as_ptr().add(index as usize * ty.layout.size());
+                (ty.drop)(removed);
             }
         }
-        self.len = last;
+        core::mem::drop(guard);
```
adamreichold: Is it actually necessary to explicit drop the guard? It will be dropped at return anyway?

tooson9010-spec: It is. The guard holds `&mut self`, and the return value reads
`self.entities[last]`, so without the explicit drop the borrow is still live:

```
error[E0503]: cannot use `self.entities[_]` because it was mutably borrowed
   --> src/archetype.rs:372:18
    |
359 |             archetype: self,
    |                        ---- `*self` is borrowed here
...
372 |             Some(self.entities[last as usize])
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ use of borrowed `*self`
```

adamreichold: You could use a scope instead of the explicit drop (which I think is a leftover and suggests the I-need-to-disarm-this-guard-after-this-operation-succeeded pattern), but this is admittedly a purely stylistic question.

tooson9010-spec: Scoped it, thanks again for your review!

Ralith:
> Sorry again for the slow reply.

You're fine, we're all volunteers here, participating as time permits.

tests/tests.rs
Ralith, line 796: Looks like the change to use thread-locals wasn't propagated back to this commit (or the next)?

src/command_buffer.rs
Ralith, line 176: I think there's still a double-drop if a component drop panics here. We should update `consumed` before attempting to move the components.

Ralith (changes requested):

src/command_buffer.rs
tooson9010-spec: Fixed. consumed is now committed before the move, so a Drop panicking inside World::insert no longer lets clear destroy the already-moved components. Added a regression test (four drops with the fix, five without); passes under Miri.

tests/tests.rs
tooson9010-spec: that's only in the earlier commits, the final tree uses thread-locals. I can rebase it back into the helper commit if you'd like the history clean, or leave it. Whichever's easier.

tests/tests.rs
```diff
@@ -889,6 +889,41 @@ fn command_buffer_does_not_double_drop_when_a_command_panics() {
...
+    drop(buffer);
+    drop(world);
+
+    // Unless `consumed` is committed before the move, the buffer's `clear`
+    // destroys the components already handed to the world a second time: the
+    // old Bomb's destructor runs once during the panic and once more here,
+    // giving five drops instead of four. Leaks are safe; a double drop is not.
+    assert_eq!(
+        drops(),
+        4,
```
Ralith, line 922: Shouldn't this be 3?

tests/tests.rs
Ralith: I'd prefer clean history, but if necessary I can squash the PR instead.

tests/tests.rs
tooson9010-spec: The fourth drop is drop(world). After insert_inner unwinds, the entity's archetype is left in a partial state and the world destructor drops the original armed Bomb again.

1. armed Bomb — insert_inner drops it while overwriting (panics)
2. Bomb { armed: false } — Reset::drain (consumed committed before the move)
3. Tracked — Reset::clear
4. armed Bomb — drop(world) ← double-drop the fix prevents

tests/tests.rs
tooson9010-spec: Please go ahead. Thanks!

tests/tests.rs
Ralith: I'm confused. If this change prevents drop 4, why are we asserting that there are 4 drops?

tooson9010-spec: Sorry for the confusion, and for making this take so long. 
I had the drop sequence wrong. 
The fix prevents Reset::drain from double-dropping Bomb { armed: false }, not drop(world). 
I thin this is the corrected sequence:

Without fix (5):
1. Bomb { armed: true } — insert_inner drops old (panics)
2. Bomb { armed: false } — world.insert drops it during unwind
3. Bomb { armed: false } — Reset::drain drops it again ← fix prevents this
4. Tracked — Reset::clear
5. Bomb { armed: true } — drop(world)

With fix (4): consumed is committed before the move, so drain skips slot 2. assert_eq!(drops(), 4) is the with-fix count.

Ralith: > Sorry for the confusion, and for making this take so long.

You're fine, there's no deadline, and I appreciate the help sorting these out! 

> I thin this is the corrected sequence

Why doesn't this constitute a double drop of the armed bomb?

tooson9010-spec: You're right, it's a double drop. Sorry, I got that wrong.

```
drop #1 armed=true  at 0x7ff7fc001240
drop #2 armed=false at 0x7ff7fc001290
drop #4 armed=true  at 0x7ff7fc001240   <- same object as #1
(Tracked doesn't print, so it isn't in this log)
```

Adding a `String` field to `Bomb` and running under Miri:

```
error: Undefined Behavior: constructing invalid value of type &mut [u8]:
       encountered a dangling reference (use-after-free)
  --> core/src/ptr/mod.rs:820   unsafe { drop_glue(&mut *to_drop) }
```

`insert_inner` drops the replaced components before writing the new ones:

```rust
// src/world.rs
for &ty in &target.replaced {
    let ptr = source_arch.get_dynamic(ty.id(), ty.layout().size(), loc.index).unwrap();
    ty.drop(ptr.as_ptr());     // unwinds here
}
```

If that unwinds, the entity is left untouched in its archetype, so dropping the
world runs the same destructor again. The new value never reaches the world and
is destroyed during the unwind (the `armed=false` line above).

So the expected count of 4 includes that second drop. Either the replaced
components get dropped after the new ones are in place, in which case the test
expects 3, or this PR stays on the command buffer and the test uses a command
that doesn't overwrite.

Ralith: If you want to limit the scope of the PR, then switching to a test which queues an insert for an entity which doesn't exist, causing the inserted components to be dropped immediately when the command buffer runs, would be an alternative test strategy which doesn't trigger this unrelated bug.

I've opened https://github.com/Ralith/hecs/issues/473 to track the new issue so we don't lose track of it. A good fix isn't immediately obvious to me, though I don't have a ton of time to think about it at the moment.

Ralith: Actually, as detailed on that issue I think there is a reasonable fix here, but we don't need to block this PR on it.

